PR #52, opened 2025-07-16
update: define n-d promotion rules for the initializing expression

JustinMeimar:
I'm not super confident this behavior is well defined for arrays of arbitrary dimension. For 2D (matrices) and 1D (arrays, formerly vectors), I believe it is. The spec clearly tries to leave the door open for N-D in the future, but examples like this are unclear: 

```
integer[2][3][4] three = [[1,2],[3,4]];
```

The initialization expression is a literal 2D array. How should that promote into a 3D array? Extrapolating from the 1D -> 2D case, the resulting contents in `three` would be:

```
[
 [ [1, 2, 0],
   [3, 4, 0]],
 [ [1, 2, 0],
   [3, 4, 0]],
 [ [1, 2, 0],
   [3, 4, 0]],
 [ [1, 2, 0],
   [3, 4, 0]],
]
```

At the end of the day Gazprea still only supports 2D for now, so this doesn't really matter, just something to think about for the future.

### DIFF
--- a/gazprea/spec/type_promotion.rst
+++ b/gazprea/spec/type_promotion.rst
@@ -83,13 +83,7 @@ operand is a square matrix (:math:`m \times m`).
 Multi-dimensional Arrays
 --------------------------
 
-Array promotion to a higher dimension occurs because, for example,
-a row in a 2D array is equivalent to a 1D array.
-When a 2D array is initialized or operated on with a 1D array, each scalar
-element in the 1D array is interpreted as a row. By applying the scalar to
-array promotion rule, each scalar element in the array will be promoted
-to a row. The example below demonstrates scalar to row promotion,
-row padding and column padding all together.
+Array promotion to a higher dimension occurs when elements of the initialization array can be promoted from the corresponding element type of the declaration. For example, when a declaration of a 2D array (matrix) is initialized with a 1D array (vector), each scalar element in the 1D array is interpreted as a row. This is because higher dimensional arrays are defined as an array of arrays, therefore each scalar element is promoted to a 1D array. The example below demonstrates scalar to array promotion and dimension padding.
 
 ::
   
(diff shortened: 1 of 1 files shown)


Sir-NoChill:
I was previously going to promote a cool method of extrapolating that the dimensionality of the slice would correspond to the least-significant dimensions, but I think that just opens the door to a whole lot of weirdness upon really examining myself.

I would argue that when (if) we extend to arbitrary dimensional arrays, this should be a dimension mismatch/size error. I think that promotion from a 1d slice literal to a higher dimensional array/vector will need to be changed when we extend:
```gazprea
procedure main() returns integer {
  // basic examples
  integer a = 2;
  integer[*] b = [a]; // type: integer[1];
  integer[*] c = a; // size error, as we have now
  integer[2][2] = a; // type: integer[2][2] = [[2, 2], [2,2]]
                     // note that here I advocate for broadcasting of padding 0
  integer[2][2] = [a]; // SizeError since elems != 4, see below
 
  // dimension deduction
  integer[4] x = [1, 2, 3, 4];
  integer[2][2] y = [1, 2, 3, 4]; // Ok. gives [[1, 2], [3, 4]];
  integer[2][3] w = [1, 2, 3, 4]; // SizeError; 6 elements should be required
  integer[2][3] q = [1, 2, 3, 4, 5, 6]; // Ok. gives [[1, 2, 3], [4, 5, 6]];
  integer[1][1][1][1][1][1] r = 1; // Ok, since the array only requires 1 element
  integer[512][2048] s = 1; // Ok because broadcasting
  integer[1][1][1][1][1][1] t = [1]; // Ok, because num_elems_lhs == num_elems_rhs
  integer[512][2048] u = [1]; // SizeError because num_elems_l != num_elems_r
  integer[*] v = [1, 2, 3]; // Ok. Normal.
  integer[*][*] z = [[1, 2, 3],[4, 5]]; // SizeError. Ragged not supported for static arrays
                                        //  INPUT WANTED: how to deal with ragged?
}
```

Do you see my idea here? As shown, I'm not entirely sure how we deal with ragged when we get there so further input is welcome.

JustinMeimar:
@Sir-NoChill  All look fine to me I think. There is one example however which demonstrates how subtle defined vector -> matrix promotion may be. For this example, it would not be an error strictly because 6 elements are not provided.

` integer[2][3] w = [1, 2, 3, 4]; // SizeError; 6 elements should be required`

If we assume row major (or a vector of row vectors), one could do:
```
integer[2][3] w = [1, 2];
```
Where each scalar in the RHS uses a `scalar -> integer[3]` promotion rule, yielding:
```
[[1, 1, 1], [2, 2, 2]]
```

The weird part is when we mix in padding and heterogeneous vectors.
```
integer[2][3] w = [1, [3,4]];   // heterogenous, 1 -> [1, 1, 1] while [3, 4] pads to [3, 4, 0]
integer[3][3] w = [1, 2];        // arguably produces [[1,1,1], [2,2,2], [0,0,0]]  
```

Does this make sense? If it's not well defined it may be better to remove matrix-vector promotion and just have broadcasting (scalar -> ndarray) which is simply defined if/when we bring ndarrays into Gazprea.

PR #89, opened 2026-02-12
WIP: Gazprea facelift based on 2025 feedback

Sir-NoChill:
Based on feedback from last year, I propose the following changes:
- A definition for constexpr
- A revised version of vectors (now contained in arrays)
  - Ragged arrays as data structures only
  - Arrays as the root of all composites in gazprea (different classes,
    e.g. dynamic, static, ragged, all handled in the backend)
  - Well defined arrays of structs
- A revised definition of tuples
  - Tuples and instantiation
  - Nested tuples
- A set of extensions for students to further experiment with

Keeping this in a seperate folder from the current specification for the
moment, but would like comments on how we might like this system to
evolve. There are rough edges that I would like to smooth out in tuples
and arrays, and a solution compiler is underway (privately)

I would appreciate comments on the current direction, specifically with
respect to tuples and arrays.


rcunrau:
That's a lot of proposals! I cannot/will not try to figure out your intent via context diffs. If you want to make proposals with illustrations of what's wrong and what you're fixing we can discuss it, but I don't think I'll have to time to discuss it before May. I also remember that we held back on a number of things like arrays of structs, nested tuples, etc. because we were concerned that they were too much work. I also remember we discussed the direction forward as towards more compiler stuff (e.g. SSA inference or dead code) and less language stuff (e.g. user defined classes/objects). Does your feedback say the students wanted more language stuff? (I heard the opposite)

PR #93, opened 2026-05-31
Gazprea 2026 changes

rcunrau:
Just a place to have a discussion. I already opened https://github.com/cmput415/415-docs/issues/92
and https://github.com/cmput415/415-docs/issues/91

### DIFF
--- a/gazprea/spec/types/array.rst
+++ b/gazprea/spec/types/array.rst
@@ -406,6 +406,7 @@ the array elements captured by the slice, as shown below.
 
 ::
 
+    // 0..10 is a range, not a slice
     integer[*] a = 0..10 by 2; /* a = [0, 2, 4, 6, 8, 10] */
     integer[2] x = a[2..4]; /* x == [2, 4] */
     integer y = a[2..4][1]; /* y == 2 */
@@ -415,8 +416,9 @@ the array elements captured by the slice, as shown below.
     integer z2 = a[1..7][1..7][1..7][4]; /* z2 == 6 */
 
 
-Array slices have special behaviour when they are used in a parameter call,
-where they allow modification of the source array:
+Array slices have special behaviour when they are used in a parameter call 
+or on the left side of an assignment, where they allow modification of the 
+source array:
 
 
 ::
@@ -427,9 +429,9 @@ where they allow modification of the source array:
 
     procedure main() returns integer {
         
-        integer[10] a = 0..10 by 2; /* a = [0, 2, 4, 6, 8, 10] */
-        integer[10] b = 0..15 by 3; /* b = [0, 3, 6, 9, 12, 15] */
-        integer[10] c;
+        integer[6] a = 0..10 by 2; /* a = [0, 2, 4, 6, 8, 10] */
+        integer[6] b = 0..15 by 3; /* b = [0, 3, 6, 9, 12, 15] */
+        var integer[6] c;          /* c must be var */
 
         /* procedure works normally with an array */
         call sum_arrays(a, b, c);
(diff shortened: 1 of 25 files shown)


rcunrau:
7.7.4 Slices - it says slices can be strided using `by`, which is not true.

The example needs to be fixed - c has 10 elements but we only print 6, and 3rd example uses `by`.

The section only talks about slices as parameters, but they can also be used in assignments and expressions:

```
integer[*] a = 1..20 by 2; // a range not a slice;
integer[*] b = a[..4]; // slice being used as an r-value 
a[8..] = 0; // slice being used as an l-value
```

We haven't talked about the 3rd example before, but I believe it is consistent with slice semantics

gazprea/spec/types/array.rst
```diff
@@ -299,7 +298,7 @@ Operations
          integer[*] a = 0..10 by 2; /* a = [0, 2, 4, 6, 8, 10] */
          integer[2] x = a[2..4]; /* x == [2, 4] */
 
-      Note that for slices only a stride of 1 is allowed.
+      Note that for slicing with this syntax always has a stride of 1.
```
rcunrau: Maybe: Note that slices always have a stride of 1?

gazprea/spec/types/array.rst
```diff
@@ -394,6 +395,64 @@ Operations
    The ``!=`` operation also produces a boolean instead of a boolean array.
    The result is the logical negation of the result of the ``==`` operator.
 
+.. _sssec:array_slices:
+
+Array Slices
+~~~~~~~~~~~~
+
+An array may be indexed by a range to create a new array that is a *slice*
```
rcunrau: Do we want to be careful about saying it "creates" a new array? I always think of a slice as <start pointer,length> even though we only use it as an r-value in the examples you have below.

gazprea/spec/types/array.rst
```diff
@@ -394,6 +395,64 @@ Operations
...
+    integer[*] a = 0..10 by 2; /* a = [0, 2, 4, 6, 8, 10] */
+    integer[2] x = a[2..4]; /* x == [2, 4] */
+    integer y = a[2..4][1]; /* y == 2 */
+
+    // A slice of the entire array behaves as the array itself, this can be repeated
+    integer z1 = a[4];                   /* z1 == 6 */
+    integer z2 = a[1..7][1..7][1..7][4]; /* z2 == 6 */
+
+
+Array slices have special behaviour when they are used in a parameter call
```
rcunrau: See? If a slice is always considered <start, length> these behaviours aren't particularly special, they are wonderfully natural :-)

gazprea/spec/types/array.rst
```diff
@@ -394,6 +395,64 @@ Operations
...
+        c -> std_output; /* [0, 5, 10, 0, 5, 10] */
+
+        /* slice can be assigned to, modifying c */
+        c[3..5] = [415, 429];
+        c -> std_output; /* [0, 5, 415, 429, 5, 10] */
+    
+        return 0;
+    }
+
+This behaviour is consistent with the slice being thought of as a
```
rcunrau, line 454: Yes! Maybe we should lead with this?

gazprea/spec/types/tuple.rst
```diff
@@ -116,7 +116,7 @@ Unpacking
 
 Any tuple expression may be assigned (unpacked) into multiple lvalues. If the size of
 the tuple being unpacked does not match the number of lvalues being asigned, an ``AssignError``
-may be raised. There is no partial unpacking of tuples.
+is raised. There is no partial unpacking of tuples.
```
rcunrau, line 119: Would we ever want to consider the Rust convention of '_' for fields we don't care about?

gazprea/spec/types/vector.rst
```diff
@@ -45,7 +47,7 @@ Those greater raise a runtime ``SizeError``.
 
         const vector<character> vec = ['a', 'b', 'c'];
         const vector<real[*]> ragged_right = [[1.0], [2.0, 2.0]]; // SizeError
-        const vector<real[*]> paddeded_right = [[1.0, 2.0], [1.0]]; // Padds second element
+        const vector<real[*]> padded_right = [[1.0, 2.0], [1.0]]; // Pads second element
```
rcunrau, line 50: Is this straight forward to implement as far as you know? In the ragged right controversy, this might be case where `real[2][*]` and `vector<real[*]> behave differently. In particular, the 2D array is dense, and therefore has to be initialized in some way. In contrast, the vector of arrays is not dense, although I guess we've concluded that working with ragged right data structures is too awkward to allow.

rcunrau (commented):
Just suggestions that we can discuss

gazprea/spec/types/vector.rst
novo52: In this case, I think both are dense, because the vector is the outer dimension. I guess `vector<real[2]>` could be a vector of array pointers depending on implementation.

With the padding and size rules, I think it is straightforward to implement.

PR #97, opened 2026-06-22
[SPEC] Constexpr Specification

Sir-NoChill:


gazprea/spec/globals.rst
```diff
@@ -24,11 +24,16 @@ place since mutable global variables would ruin functional purity.
...
+Globals must be initialized with a valid
+:ref:`constant expression <sec:constexpr>`. A global initializer may therefore
+reference other globals and use arithmetic and constexpr aggregates, but it must
+be fully evaluable by the compiler before the program runs. This preserves
+functional purity and enables compile-time optimizations. As a consequence:
+
+*   Functions, procedures, and I/O operations may not appear in a global's
+    initializer.
+*   A global may not have a dynamically-sized array type (e.g. ``integer[*]``),
+    because its size would not be known at compile time.
```
novo52:
```suggestion
```
Not true: `integer[*] X = [1,2,3];` is known.
Conflation of inferred size and dynamic size; will be confusing without clarification here or elsewhere.

gazprea/spec/constexpr.rst
```diff
@@ -0,0 +1,143 @@
...
+
+.. _ssec:constexpr_rules:
+
+Rules for Constant Expressions
+------------------------------
+
+An expression is a valid ``constexpr`` if it is composed exclusively of:
+
+1.  Literals of base types (``boolean``, ``integer``, ``real``, ``character``).
+2.  Operators, including ``+``, ``-``, ``*``, ``/``, ``not``, ``and``, ``or``,
```
novo52:
```suggestion
2.  Operators ``+``, ``-``, ``*``, ``/``, ``not``, ``and``, ``or``,
```
"Exclusively" does not work with "including". I assume these are the only valid operations.

gazprea/spec/globals.rst
```diff
@@ -24,11 +24,16 @@ place since mutable global variables would ruin functional purity.
...
-time to run time.
+Globals must be initialized with a valid
+:ref:`constant expression <sec:constexpr>`. A global initializer may therefore
+reference other globals and use arithmetic and constexpr aggregates, but it must
+be fully evaluable by the compiler before the program runs. This preserves
+functional purity and enables compile-time optimizations. As a consequence:
+
+*   Functions, procedures, and I/O operations may not appear in a global's
+    initializer.
+*   A global may not have a dynamically-sized array type (e.g. ``integer[*]``),
```
novo52:
```suggestion
*   A global may not have a inferred-sized array type (e.g. ``integer[*]``).
```

novo52 (changes requested):
typedef.rst says an array-alias size may use only "arithmetic operations on scalar literals … not constant propagation." constexpr rule 5 explicitly permits "Other variables that are themselves valid constexprs," and the array example sizes with integer[WIDTH] / integer[ELEMENT]. So typealias integer[WIDTH] foo; is illegal per typedef.rst but should be legal under the constexpr model.
My suggestion: array-alias sizes should be constexpr

array.rst does not speak on statically-sized or dynamically-sized arrays at all. These are different from explicit vs inferred size. Are explicitly sized arrays necessarily statically-sized? See the example on line 121, this is an explicit-and-dynamically-sized array. Clarify either here or in arrays.rst.

gazprea/spec/globals.rst
Sir-NoChill: Apologies, these are artifacts from my vector/array proposals. Updating accordingly, thank

gazprea/spec/constexpr.rst
Sir-NoChill: Do you consider an expression like `var a <- std_input` to have the `<-` operator? I think this is what I'm gating against. Maybe a better wording is `The operators ``+``, ``-``, ...`

gazprea/spec/globals.rst
Sir-NoChill: However a global may have an inferred sized array type, we just need to make sure that the array type size is inferrable at compile time (eek that sounds circular). I can't think of a way this does not hold in the current spec.

gazprea/spec/constexpr.rst
```diff
@@ -77,13 +77,16 @@ allowing them to be used to define other constants.
...
+   A ``const`` array is a ``constexpr`` if:
 
    1. Its size is a valid ``constexpr``.
    2. All of its element initializers are valid ``constexpr``\ s.
 
-   Dynamically-sized arrays (e.g., ``integer[*]``) cannot be ``constexpr``
-   aggregates as their size is not known at compile time.
+   A ``vector`` (the dynamically-sized type) can never be a ``constexpr``
+   aggregate, since its size is determined at runtime. An inferred-size array
+   such as ``integer[*] X = [1, 2, 3]`` *can* be a ``constexpr`` when its
```
novo52: I feel "can" is the wrong language--an inferred size array must be `constexpr`, as all arrays must be. Therefore, for `[*]`, the RHS must be `constexpr`.

gazprea/spec/constexpr.rst
novo52: Yes, that is good. The problem with "including" is that it implies other operators can also be `constexpr`, but (I believe) your intent is to communicate that this is an exhaustive list of `constexpr '- compatible operators.

novo52 (approved):
With the two changes I suggested, this is overall an excellent change. I think that making arrays static in size is a win for student learning. They will have the opportunity to do optimizations that would be complicated by dynamically-sized arrays, and the distinction between vectors and arrays allows student compilers an "easy-out" if they do not wish to do the same optimizations for vectors; they can implement vectors with {cap, len, ptr}, but arrays are an easy optimization win for them.

gazprea/spec/constexpr.rst
Sir-NoChill: good catch

gazprea/spec/types/array.rst
```diff
@@ -26,9 +26,11 @@ array instead of a ``real`` array.
 
 #. Explicit Size Declarations
 
-   When an array is declared it may be explicitly given a size. This
-   size can be given as any integer expression, thus the size of the
-   array may not be known until runtime.
+   When an array is declared it may be explicitly given a size. This size
+   must be a :ref:`constant expression <sec:constexpr>`. Every array, whether
```
rcunrau: I disagree with this proposal

gazprea/spec/constexpr.rst
```diff
@@ -0,0 +1,146 @@
...
+
+An expression is a valid ``constexpr`` if it is composed exclusively of:
+
+1.  Literals of base types (``boolean``, ``integer``, ``real``, ``character``).
+2.  The operators ``+``, ``-``, ``*``, ``/``, ``not``, ``and``, ``or``,
+    between two or more ``constexpr``\ s.
+3.  Constructors for aggregate types, provided that the aggregate is const and
+    all members are ``constexpr``\ s.
+4.  Index or field access on ``constexpr`` aggregate types.
+5.  Other variables that are themselves valid ``constexpr``\ s.
```
rcunrau, line 28: I guess we agreed that literals only was too restrictive? I thought the reason we were restricting to literals only was so the teams did not have to do constant propagation?

rcunrau: Do I remember you saying that the solution is to store the constants values in the symbols, and that you were going to give them the MLIR tool to do symbolic expression evaluation?

rcunrau (changes requested):
Do we still have an implementation section? Perhaps you can include a pointer to the MLIR expression evaluation tool there (sorry if it's there already and I missed it)

gazprea/spec/types/array.rst
Sir-NoChill: I'm not sure I understand... this is just saying that arrays are statically sized in my mind, and that the allocation size does not need to be determined at runtime?

We can say it has to be an integer, if that's the sticky point.

gazprea/spec/constexpr.rst
Sir-NoChill: So what happens right now is that all of this is automatically taken care of by constant propagation, anything needing constexprs that is not directly computable via constant propagation is done in a ctors block in the IR.

gazprea/spec/types/array.rst
rcunrau: Perhaps I am confused: I thought the definition of a `constexpr` is that it can be evaluated at compile time?

It should be that gazprea allows:

```
integer size = random() & 0x255;
integer[size] surprise;
```

I don't find the wording that you used is super clear that this is OK

gazprea/spec/constexpr.rst
rcunrau: Suppose we have this:

```
const integer x = 5;
const integer y = x * 2;
const integer z = 10.0/0.0;
```

Does `y` need a `ctor`? I can see how I could store `x` as a literal and do my own constant propagation in my AST. But the math part is trickier, in that it seems like you have to generate the MLIR expression and then see if it can be simplified. Ideally `z` could be evaluated as a compile time error but it actually takes a lot of work!

gazprea/spec/types/array.rst
Sir-NoChill: Hmm, okay, I think I see. That would make them all 'variable-length arrays' if we steal the C99 term meaning that the size is bound once, upon [elaboration](https://aada.m2osw.com/alexis-ada-reference-manual/annex-n-glossary-informative#elaboration)?

This breaks alignment with rust-like semantics, as long as you're okay with that:

```rust
use rand::prelude::*;

fn main() {
    let /* cannot declare const */ a: u8 = rand::random();
    let c = [0; a]; // error: a is non-constant
    println!("{:?}", c);
}
```

gazprea/spec/constexpr.rst
Sir-NoChill: Yeah, I currently have a miniature interpreter embedded in sema to evaluate constexprs. So the current implementation would emit a 'global error' without nice errors enabled, if you go through the errors then you'll see that `z` fails to evaluate since 10.0/0.0 is invalid (maybe it should be a nan, but this is registered as a type of global error for now).

If we change it to be legal (`const integer z = 10.0`) then we get the following mlir:

```mlir
module {
  gaz.global @x : i32
  gaz.global @y : i32
  gaz.global @z : i32

  gaz.init { // the ctor block
    %c5_i32 = arith.constant 5 : i32
    gaz.set_global @x, %c5_i32 : i32
    %c2_i32 = arith.constant 2 : i32
    %0 = gaz.binary mul %c5_i32, %c2_i32 : (i32, i32) -> i32
    gaz.set_global @y, %0 : i32
    %cst = arith.constant 1.000000e+01 : f32 // If I attempt to divide by zero here, we catch arithmetic in sema and throw a global error that it is not a constexpr since it cannot evaluate
    %1 = gaz.cast %cst : f32 to i32
    gaz.set_global @z, %1 : i32
  }
```

The other way to do this is to have the expressions fully analyzed in sema (so our mini interpreter) and then attach them to the globals as attributes, but I have not played with that yet.

All of that said, the students can leave all of this to mlir, they do not need to implement any constant propagation themselves, it just makes the code generation a bit messier at first. I am doing it this way to make the IR pretty.

Doing it a different way you could declare the globals and then have all of the value definitions generated in main. LLVM constant propagation would lower it correctly and evaluate the constants when we lower to LLVM IR, what I did was hoist it to its own special block to add the constraint that globals cannot be changed from anywhere else in the program (set_global must have gaz.init as its parent).

gazprea/spec/types/array.rst
Sir-NoChill: This is updated.

PR #100, opened 2026-07-16
Added advice for students

novo52:
I want the students to understand certain decisions that will affect them down the line when implementing their compilers.

I am a bit iffy on how it is presented. I appreciate criticism.

### DIFF
--- a/gazprea/impl/backend.rst
+++ b/gazprea/impl/backend.rst
@@ -6,6 +6,163 @@ Backend
 You don’t need to implement an interpreter for Gazprea. You only need to
 implement a *MLIR* code generator that outputs *LLVM IR*.
 
+.. _ssec:representing_values:
+
+Representing Values
+-------------------
+
+When you emit MLIR you must decide *how a value lives*: as an SSA value that is
+defined once, or as a slot in memory that you load from and store to. Pick one deliberately.
+
+.. note::
+
+   **In practice.** The memory model (``alloca`` / ``load`` / ``store`` +
+   ``mem2reg``) is standard for imperative, systems-style languages with an
+   explicit, language-defined memory layout, like C. Value-semantic ``tensor``
+   representations are favored at the high level by array- and math-focused
+   compilers like Fortran and Tensorflow.
+
+**The memory model.** Give every value a slot: reads are loads, writes are stores.
+The mutable state lives in memory and the optimizer's ``mem2reg`` promotion turns the slots back into SSA
+registers for you. A scalar declaration is an ``alloca``, a read a ``load``, an
+assignment a ``store``:
+
+::
+
+   // var integer n = 42;  then  n = n + 1;
+   %n = memref.alloca() : memref<i32>
+   memref.store %c42, %n[] : memref<i32>
+   %0 = memref.load %n[] : memref<i32>
+   %1 = arith.addi %0, %c1 : i32
+   memref.store %1, %n[] : memref<i32>
+
+It pairs naturally with unstructured ``cf``: because the state is in memory,
+``break`` / ``continue`` / ``return`` are ordinary branches — to the loop's exit
+block, its latch, and the function exit — with nothing threaded through them and
+no analysis of which variables cross a construct. 
+
+**The value (SSA) model.** Map each variable name to its *current* SSA value: a
+read is a lookup, an assignment produces a new value and rebinds the name. There
+are no slots and no loads or stores; bufferization introduces memory later,
+downstream of your emitter. The one hard case is a *merge*: after an ``if``, a
+variable set on only one branch still needs a single value afterward. The
+structured ``scf`` ops hand you this by *yielding* merged values as region results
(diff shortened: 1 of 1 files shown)


gazprea/impl/backend.rst
```diff
@@ -6,6 +6,163 @@ Backend
...
+to its ``scf.yield`` — so Gazprea's ``break``, ``continue``, and ``return`` need a
+way to be represented *inside* a structured region; work out how before you commit
+to this model.
+
+.. Warning::
+   Avoid emitting unstructured ``cf`` blocks and
+   placing the phi / block-argument merges yourself.
+
+.. Warning::
+   The SSA model is the more difficult of the two to implement, but can allow for more
```
rcunrau: Do you know for a fact that more optimizations can occur? I might be persuaded myself if you have a demonstrable example. My guess is that LLVM has a pass that does this for you, so that it can catch all the opportunities.

JustinMeimar (approved):

gazprea/impl/backend.rst
```diff
@@ -6,6 +6,80 @@ Backend
...
 implement a *MLIR* code generator that outputs *LLVM IR*.
 
+.. _ssec:representing_values:
+
+Representing Values
+-------------------
+
+When you emit MLIR you must decide *how a value lives*: as a slot in memory that
+you load from and store to, or as an SSA value that is defined once. **Use the
+memory model.** It keeps code generation local and mechanical. The choice is the
```
Sir-NoChill, line 20: I'm not sure we should outright suggest this. There is a good argument for using value-typed systems that FLang uses in FIR, namely they associate an fir expression that can grow and shrink (above memref) and then bufferize down to fir.box and fir.ref types with allocations. There is also precedent for this in the sparse_tensor dialect, which has a 'push_back' op that creates a new value type whenever adding an element to an array.

I agree that if the students were to implement this all themselves it just results in a ton of extra work, but they can steal from other dialects to make it somewhat easier. Especially with the bots, adopting the FIR/CLangIR methods would be effective in my mind.

gazprea/impl/backend.rst
```diff
@@ -6,6 +6,80 @@ Backend
...
+
+.. Note::
+   For performance marks, consider how your vector grows.
+
+The value (SSA) model
+~~~~~~~~~~~~~~~~~~~~~
+Understand the formulation, avoid implementing it.
+
+*Scalars.* Map each variable name to its *current* SSA value: a read is a lookup, an
+assignment produces a new value and rebinds the name. There are no slots and no
```
Sir-NoChill, line 76: Might be useful to link to the bufferization documentation, since this comes with a little bit more nuance when you look at the bufferization passes: https://mlir.llvm.org/docs/Bufferization/#destination-passing-style

We may also want to discuss the composite types here too, and how MLIR can handle the destructuring for the students automatically. Probably pointing to toy ch7

gazprea/impl/backend.rst
```diff
@@ -22,12 +96,12 @@ Below is an example of how to use ``malloc`` and ``free`` within MLIR using the
...
-    llvm.func @free(!llvm.ptr<i8>)
+    llvm.func @malloc(i64) -> !llvm.ptr
+    llvm.func @free(!llvm.ptr)
     llvm.func @main() -> i32 {
-      %0 = llvm.mlir.constant(128 : i32) : i32
-      %1 = llvm.call @malloc(%0) : (i32) -> !llvm.ptr<i8>
-      llvm.call @free(%1) : (!llvm.ptr<i8>) -> ()
+      %0 = llvm.mlir.constant(128 : i64) : i64
+      %1 = llvm.call @malloc(%0) : (i64) -> !llvm.ptr
+      llvm.call @free(%1) : (!llvm.ptr) -> ()
```
Sir-NoChill: I don't think the students should _prefer_ to use the llvm dialect for the most part. In mlir this is supposed to be treated as a target dialect, rather than a dialect in which you write your IR, otherwise you just end up writing LLVM IR since MLIR cannot optimize LLVM IR (see the output dialects). I think we should prefer using func, memref and arith for this.

Sir-NoChill (changes requested):
Looking good, but I think we should be deliberate about suggesting that the students use the tools that a dialect comes with by default to get some of the productivity, optimization and stylistic gains that MLIR can provide. I left some comments, I'd like to discuss them.

gazprea/impl/backend.rst
Sir-NoChill: Could also point to the pytorch dialect that has the NonValueTensorType which might be a useful model. It exists above bufferization but is also mutable, I think that's the model we should follow in the solution.

novo52: I think for this year, we should hold off on these types of improvements so the feedback from the change in evaluation is less noisy.

Sir-NoChill:
What are our thoughts here?

rcunrau:
My opinion has not changed: this belongs in a formal lecture, not in the appendix of a spec. We can leave it open until we do something more permanent with it, or I suppose it will still be here if we park/close it.

PR #102, opened 2026-07-28
Lolcode spec fixes

novo52:
Some fixes/clarifications to the lolcode spec

### DIFF
--- a/lolcode/spec/expressions.rst
+++ b/lolcode/spec/expressions.rst
@@ -21,18 +21,6 @@ Math is performed as integer math in the presence of two ``NUMBR`` s.
 If either of the arguments is a ``YARN``, the ``YARN`` is
 converted to an integer and the operation proceeds if the conversion succeeds.
 
-Conditions
-~~~~~~~~~~
-
-LOLCODE supports two conditional operators:
-
-::
-
-    BIGGR OF <x> AN <y>     BTW max
-    SMALLR OF <x> AN <y>    BTW min
-
-These operators apply to both strings and integers. As with the math operators above, if either argument is a ``YARN``, it is converted to an integer and the operation proceeds if the conversion succeeds.
-
 Concatenation
 ~~~~~~~~~~~~~
 
@@ -40,4 +28,3 @@ An indefinite number of YARNs may be explicitly concatenated with the ``SMOOSH..
 Arguments may optionally be separated with ``AN``.
 As the ``SMOOSH`` expects strings as its input arguments, it will implicitly cast all input values of other types to ``YARN`` s.
 The line ending may safely implicitly close the ``SMOOSH`` operator without needing an ``MKAY``.
-
--- a/lolcode/spec/expressions.rst
+++ b/lolcode/spec/expressions.rst
@@ -31,8 +31,7 @@ LOLCODE supports two conditional operators:
     BIGGR OF <x> AN <y>     BTW max
     SMALLR OF <x> AN <y>    BTW min
 
-These operators apply to both strings and integers. Since both types are
-supported, the types of ``<x>`` and ``<y>`` must match.
+These operators apply to both strings and integers. As with the math operators above, if either argument is a ``YARN``, it is converted to an integer and the operation proceeds if the conversion succeeds.
 
 Concatenation
 ~~~~~~~~~~~~~
--- a/lolcode/spec/expressions.rst
+++ b/lolcode/spec/expressions.rst
@@ -21,17 +21,6 @@ Math is performed as integer math in the presence of two ``NUMBR`` s.
 If either of the arguments is a ``YARN``, the ``YARN`` is
 converted to an integer and the operation proceeds if the conversion succeeds.
 
-Precedence
-~~~~~~~~~~
-
-Multiplication, division and modulo have the highest precedence. All three
-operators have equal precedence and are evaluated left to right.
-Addition and subtraction have equal precedence and are also evaluated left to
-right.
-
-Parenthesis ('()') can be used to group operations to override precedence or
-clarify the evaluation order.
-
 Conditions
 ~~~~~~~~~~
 
--- a/lolcode/spec/statements.rst
+++ b/lolcode/spec/statements.rst
@@ -16,6 +16,8 @@ Our version of LOLCODE supports seven types of statements:
 If-then
 ~~~~~~~
 
+A bare expression statement (one not assigned to a variable) stores its result in the implicit variable ``IT``. ``IT`` holds that value until the next bare expression replaces it, and is what ``O RLY?`` and ``WTF?`` test against.
+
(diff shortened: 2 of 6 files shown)


lolcode/spec/expressions.rst
```diff
@@ -21,17 +21,6 @@ Math is performed as integer math in the presence of two ``NUMBR`` s.
...
-~~~~~~~~~~
-
-Multiplication, division and modulo have the highest precedence. All three
-operators have equal precedence and are evaluated left to right.
-Addition and subtraction have equal precedence and are also evaluated left to
-right.
-
-Parenthesis ('()') can be used to group operations to override precedence or
-clarify the evaluation order.
-
```
novo52, line 34: precedence is implicit because of the operators

lolcode/spec/variables.rst
```diff
@@ -7,6 +7,8 @@ Scope
 All variable scope is local to the main program block.
 Variables are only accessible after declaration, and there is no global scope.
 
+Blocks introduced by ``O RLY?``/``YA RLY``/``MEBBE``/``NO WAI``, ``WTF?``/``OMG``, and ``IM IN YR``/``IM OUTTA YR`` are their own scope: a variable declared with ``I HAS A`` inside one of these blocks is only visible within that block, and ceases to exist once the block ends.
```
novo52: I assume this is the intention, we may also want a single global scope? However, the 
```
IM IN YR <label> <operation> YR <variable> [TIL|WILE <expression>]
  <code block>
IM OUTTA YR <label>
```
loop has a temporary local variable, so they have to do scoping anyway.

lolcode/spec/variables.rst
```diff
@@ -15,13 +17,13 @@ Variable identifiers may be in all uppercase or lowercase letters (or a mixture
 Types
 ~~~~~
 
-For the purposes of this assignment, LOLCODE only recognizes three types:
+For the purposes of this assignment, LOLCODE recognizes four types, though only three of them can be declared as variables:
```
novo52, line 21: This is like SCalc

lolcode/spec/expressions.rst
```diff
@@ -31,8 +31,7 @@ LOLCODE supports two conditional operators:
     BIGGR OF <x> AN <y>     BTW max
     SMALLR OF <x> AN <y>    BTW min
 
-These operators apply to both strings and integers. Since both types are
-supported, the types of ``<x>`` and ``<y>`` must match.
+These operators apply to both strings and integers. As with the math operators above, if either argument is a ``YARN``, it is converted to an integer and the operation proceeds if the conversion succeeds.
```
rcunrau: I was going from the compare op in the spec:

```
Comparisons are performed as integer math in the presence of two NUMBRs, but if either of the expressions are NUMBARs, then floating point math takes over. Otherwise, there is no automatic casting in the equality, so BOTH SAEM "3" AN 3 is FAIL.
```

lolcode/spec/statements.rst
```diff
@@ -16,6 +16,8 @@ Our version of LOLCODE supports seven types of statements:
 If-then
 ~~~~~~~
 
+A bare expression statement (one not assigned to a variable) stores its result in the implicit variable ``IT``. ``IT`` holds that value until the next bare expression replaces it, and is what ``O RLY?`` and ``WTF?`` test against.
```
rcunrau: `IT` is used in a bunch of places in the spec, but not anywhere else in *our* spec. Do you think we should introduce it?

lolcode/spec/variables.rst
rcunrau: I thought I had a **Scope** section that reads something like:

```
All variable scope, as of this version, is local to the enclosing function or to the main program block. Variables are only accessible after declaration, and there is no global scope.
```

lolcode/spec/variables.rst
rcunrau: SCalc (and VCalc) only have integers. I found the interpreter was *much* trickier to write with two types (and automatic promotion). It actually made we want to do more passes (symbol creation and type checking) before doing the interpreter.

lolcode/spec/expressions.rst
rcunrau: I added the parenthesis because I didn't understand how precedence works. Can you show me some examples? Can we put them in the spec so it's clear for the students?

rcunrau: I might get it: is `x+2*y` written as `sum of x an produkt of 2 an y`?

rcunrau (commented):
Good analysis and good suggestions!

lolcode/spec/variables.rst
novo52: Yes, this is that section. However, 
1. This version does not have functions
2. `<variable>` in the loop example above is local according to the spec, so your scope rule alone is inconsistent with this part of the spec.

I would prefer that we have no scope, as SCalc has no scope and I have written VCalc exams with the assumption that VCalc introduces scoping. However, IDK how to resolve this—maybe the loop requires an already declared variable?

lolcode/spec/variables.rst
novo52: When I say "this is like SCalc", I mean that SCalc also has an implicit boolean type that exists for conditionals but can not be explicitly created by the programmer.

lolcode/spec/expressions.rst
novo52: Yes, exactly. And `(x+2)*y` is written as `produkt of sum of x an 2 an y`

lolcode/spec/variables.rst
rcunrau: My reading of the Scope section is that there is no scope since we don't have functions, i.e. it already was like SCalc. Somehow I thought you were adding block scoping with your suggested changes.

I suggest we resolve the weird loop iterators exactly as intended by the spec: they are special variables that do not need to be declared and only exist within their loop body. That's the great fun of inventing your own language - you can do whatever you want!

novo52: That is what I was saying, but I no longer hold that position.

I am thinking of how the student implement the parser/interpreter. If there are these temporary variables that only exist within the loop body, then they must implement a scope table—which they did not have to do for SCalc. Then, lolcode does not have global scope, it also has block scoping rules, but only for this one specific part of the language.

I think this is not educational for the students, and we should either do real block scoping (not recommended) or get rid of this specific type of block scoping (recommended).

lolcode/spec/statements.rst
novo52: Decision: Remove `IT`, use explicit conditions in every case

lolcode/spec/variables.rst
novo52: Decision: advise students they should not bother with a scope table, all global scope except for implicit loop vars. 

Note: investigate nested loops

PR #103, opened 2026-07-28
Add architecture properties to implementation guidelines

novo52:
Gazprea gains six inspected properties alongside the existing pass list; VCalc gains a note on which of its assumptions do not survive the transition.

Also fixes a missing word in both pass-list bullets and an unindented paragraph under Software Engineering Process that warned on build.

### DIFF
--- a/info/grading.rst
+++ b/info/grading.rst
@@ -50,7 +50,7 @@ Code Style and Consistency
 * You are expected to separate class definitions from implementations using header (.h) and source (.cpp)
   files.
 * Your code should be clean and readable.
-* There is no minimum expectation for commenting or documentation.
+* There is no minimum expectation for commenting or documentation, except where the implementation guidelines require a design decision to be recorded.
 
 Software Engineering Process (Gazprea only)
 ---------------------------------------------------
@@ -59,8 +59,7 @@ Software Engineering Process (Gazprea only)
 
 * Each team should be ready to present evidence that they adhered to their plan. It is suggested to create a `PLAN.md` (or similarly named) file in the base of your repo. Updating your plan throughout the project is fine so long as it represents an improvement to the process.
 
-* Failure to specify a thorough plan or to adhere to it may result in mark deductions from this category.
-Otherwise, this should be an easy 5%. The most successful teams will establish a clear software engineering process even without being explicitly required.
+* Failure to specify a thorough plan or to adhere to it may result in mark deductions from this category. Otherwise, this should be an easy 5%. The most successful teams will establish a clear software engineering process even without being explicitly required.
 
 Grammar
 ---------------------------------------------------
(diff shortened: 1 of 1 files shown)


rcunrau (approved):

info/grading.rst
```diff
@@ -104,6 +105,20 @@ Implementation Guidelines Compliance
...
+      * **One source of truth for conversions.** The rules deciding whether a conversion is legal and the code emitting that conversion must not be two lists kept in agreement by hand. If they are separate, something in your build must check that they agree.
+
+      * **Pass dependencies are written down.** Each pass must state what it requires to already be true when it runs. If reordering two of your passes breaks your compiler, that dependency must appear somewhere a reader can find it.
+
+      * **Element-wise operations share their emission.** Adding a new operator over vectors or matrices must not require writing new index arithmetic.
+
+      * **Names are resolved once.** Your emission pass must not look a name up by string. Symbol resolution happens in an earlier pass, and later passes use the resolved symbol.
+
+      * **Locations are recorded at construction.** Every node carries the source location it came from, assigned when the node is built.
+
```
Sir-NoChill, line 121: Have we given you access to historical gazprea repos? If not I think something useful may be to run a claude over the rest of the gazprea repos to see where each of them did well or maybe made subpar design decisions. I think this is a good start, and I like having this as explicit markable criteria.

info/grading.rst
```diff
@@ -50,7 +50,7 @@ Code Style and Consistency
 * You are expected to separate class definitions from implementations using header (.h) and source (.cpp)
   files.
 * Your code should be clean and readable.
-* There is no minimum expectation for commenting or documentation.
+* There is no minimum expectation for commenting or documentation, except where the implementation guidelines require a design decision to be recorded.
```
Sir-NoChill, line 53: Maybe there should be? One thing I found useful was doc comments for major methods, and I know this has become a matter of 🤖 _taste_ 🤖 but I think it might be important to discuss how the new bot-driven development paradigm might benefit from better code documentation (but not too much!)

Sir-NoChill (commented):
I like this as well. We might want to ask around the lab allumni for what their jobs are asking for in terms of code quality as well. Maybe put out the question to nelson and see if he can get in touch with our AMD- and NVIDIA-employed alumni.

info/grading.rst
rcunrau: Since the Parsing assignment will not have interviews, I am thinking that the students will have to explain their design. There are two things I think could be marked: the grammar and the AST design. I think I found that any old grammar can work, but writing a good grammar makes the parsing easier. Same for the AST - a bad one makes it awkward to get a working interpreter.

novo52:
superseded by #137

PR #105, opened 2026-07-31
[FIX]: Gazprea iterators are non-mutable within their nested scope

Sir-NoChill:
This is my interpretation of the spec, that for iterator values, it is invalid to assign to them in the iterator body,  but that variables can shadow the inner definition. Is that correct?

The second fix is for loop iterators. I have this implemented such that if you shadow the iterator, you effectively hide the iterator variable (as per usual with nested scopes) and then at the start of the next loop before redefinition the iterator variable takes on the iterator value.

### DIFF
--- a/gazprea/spec/expressions.rst
+++ b/gazprea/spec/expressions.rst
@@ -144,8 +144,11 @@ instance:
 
 This is true for domain expressions within generators as well.
 
-Iterator variables can be assigned to and re-declared within the enclosed iterator loop.
-The variable is re-initialized according to the expression each iteration.
+An iterator variable is a fresh *constant* binding in each iteration:
+assigning to it is an error, but a declaration inside the loop body may
+shadow it, exactly as in any other nested scope. The shadowing declaration
+does not affect the iteration; the iterator is re-bound from the domain at
+the start of each iteration.
 
 ::
 
--- a/gazprea/spec/statements.rst
+++ b/gazprea/spec/statements.rst
@@ -359,8 +359,10 @@ Array ranges can also be used instead:
              i -> std_output;
            }
 
-The domain is evaluated once during the first iteration of the loop. Each iteration
-defines a constant domain variable from it's respective index. For instance:
+The domain is evaluated once during the first iteration of the loop. Each
+iteration defines a constant domain variable from its respective index;
+the variable cannot be assigned to, though it may be shadowed by an inner
+declaration (see :ref:`ssec:expressions_dom_expr`). For instance:
 
 ::
 


rcunrau (changes requested):
Just to be formal, I disagree with "unnecessary" changes to the language.

gazprea/spec/expressions.rst
```diff
@@ -144,8 +144,11 @@ instance:
 
 This is true for domain expressions within generators as well.
 
-Iterator variables can be assigned to and re-declared within the enclosed iterator loop.
-The variable is re-initialized according to the expression each iteration.
+An iterator variable is a fresh *constant* binding in each iteration:
+assigning to it is an error, but a declaration inside the loop body may
+shadow it, exactly as in any other nested scope. The shadowing declaration
+does not affect the iteration; the iterator is re-bound from the domain at
+the start of each iteration.
```
Sir-NoChill, line 151: I think that the following example would be legal for conforming implementations in the past, since the following example
```rust
for i in iter {
  i = some_var; }
```
would have a scope trace like `{scope} -> for scope (iter 1) -> {scope} -> for scope (iter 2)` which would obliterate the first definition of i in the first iteration scope. That means that `i` is reinitialized at every iteration in the loop. So just to make the spec clear we could say

> Iterator variables are re-initialized for each iteration of their corresponding loop. They may be reassigned and re-declared within the enclosed iterator loop. The variable is re-initialized to the corresponding iterator-expression upon each iteration.
>
> The iterator follows similar rules. If the user re-declares or re-assigns the iterator, it is re-initialized to the original loop iterator at the beginning of every iteration.

gazprea/spec/statements.rst
```diff
@@ -359,8 +359,10 @@ Array ranges can also be used instead:
              i -> std_output;
            }
 
-The domain is evaluated once during the first iteration of the loop. Each iteration
-defines a constant domain variable from it's respective index. For instance:
+The domain is evaluated once during the first iteration of the loop. Each
+iteration defines a constant domain variable from its respective index;
+the variable cannot be assigned to, though it may be shadowed by an inner
+declaration (see :ref:`ssec:expressions_dom_expr`). For instance:
```
Sir-NoChill, line 365: Here I will say that:

> The domain is evaluated once during the first iteration of the loop. Each iteration defines a domain variable from its respective index, for loop evaluation see :ref:`sec:expressions`.

gazprea/spec/expressions.rst
rcunrau: I think I stop worrying about all this once I actually implemented the iterator in llvm. The code I generated was something like:

```
_lower = 0;
_upper = iter.len();
for (_index=_lower; _index<_upper; _index++) {
    i = iter[_index];
    [rest of body]
}
```

If you implement the loop this way, and honestly I don't think there's too many other ways to implement it, you get exactly the semantics we describe, i.e. do whatever you want to `I` in the body, we will reinitialize it at the top of the loop.

PR #106, opened 2026-07-31
[SPEC]: Vector Semantics?

Sir-NoChill:
I am not entirely sure I understand the current reading of vectors and would like to clarify it. I have a type lattice proposed that looks as follows:

type_lattice_proposed

Does this align with what you are going for? Are arrays and vectors able to be used interchangably? Can a vector be used as an iterator? If so what happens if the vector is mutated inside the iterator scope? Do we need to COW it?

Procedures returning vectors? Procedures taking vectors as arguments makes sense but can functions take vectors as arguments?

### DIFF
--- a/gazprea/impl/errors.rst
+++ b/gazprea/impl/errors.rst
@@ -177,6 +177,8 @@ Here is an example invalid program and a corresponding compile-time error:
 
     ReturnError on line 1: procedure "main" does not have a return statement reachable by all control flows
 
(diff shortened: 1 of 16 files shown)


rcunrau:
Not sure what you want to do with all this. It's way too much for the students but it seems it should go somewhere. Have you shown it to Nelson? Perhaps you could work it into a paper, or adopt it into the grad version of the course?

PR #108, opened 2026-08-03
[SPEC]: add spec glossary with authoritative citations and cross-link student-facing pages

Sir-NoChill:
## Summary

Introduces a new [glossary page](gazprea/spec/glossary.rst) for the Gazprea specification and refactors student-facing pages to cross-reference it via Sphinx `:term:` roles instead of the current ad-hoc terminology (\"invalid\", \"illegal\", \"runtime\", \"compile-time\", \"base type\", \"l-value\", \"r-value\", \"pure function\", \"aggregate type\", ...).

Every glossary entry carries a primary citation to an authoritative source: an ISO/IEC or IEEE standard, an LLVM/GCC/Binutils manual, or a peer-reviewed publication. Where a term has a widely-used *effective* reference (cppreference for the C++ value categories), it is included alongside the normative citation and explicitly labelled as non-normative. A **Rejected sources** appendix records what was considered and passed over, so a reviewer can audit the source-selection choices.

## Commit series (5 commits, all auditable)

1. **`docs(glossary): add spec glossary with authoritative citations`** — adds `spec/glossary.rst` and registers it in the top-level toctree. Self-contained; no other files touched.
2. **`docs(spec): link student-facing pages to glossary terms`** — first-wave terminology sweep across the pages the glossary directly covers (comments, constexpr, declarations, expressions, functions, globals, identifiers, namespaces, procedures, statements, streams, type_casting, type_inference, type_promotion, type_qualifiers, typedef, types/{array,struct,tuple,vector}). Also rewords two external hyperlinks (Ada elaboration wiki, Wikipedia lvalue/rvalue) as `:term:` references so the spec is self-contained.
3. **`docs(glossary): add iterator, domain, and re-initialization terms`** — adds six related terms (`iterator`, `iterator variable`, `domain`, `domain expression`, `re-initialization`, `re-declaration`) with cites to Ada RM 5.5, C++ `[iterator.requirements]`, C++ `[stmt.ranged]`, and Python PEP 234, plus the three new citation footnotes. Fixes the vocabulary gap between statements.rst and expressions.rst that a recent iterator-mutability PR discussion had to work around.
4. **`docs(spec): link built-ins and type pages to glossary terms`** — second-wave sweep across `built_in_functions.rst` and per-type pages (character, matrix, string) that the initial patch did not touch. Files with no natural glossary sites (`keywords.rst`, `types.rst`, `types/{boolean,integer,real}.rst`) are deliberately left unchanged.
5. **`docs(globals): rewrap line-broken :term: reference`** — pure cleanup: rewraps a `:term:\`functional purity\`` reference that had been split across two source lines during the initial patch application. Sphinx renders it fine either way, but a single-line role is easier for grep tooling to inspect.

## Test plan

- [x] `sphinx-build -n -E -W --keep-going -b html . _build/html` on the whole `gazprea/` tree — clean build, no warnings, no errors (nitpicky mode with warnings-as-errors).
- [x] Every `:term:` reference across the tree resolves to a defined glossary entry (60 terms, 0 unresolved).
- [x] Every `:ref:` reference across the tree resolves to a defined label (119 labels, 0 unresolved).
- [x] Reviewer sanity-check: skim the rendered glossary page for citation formatting and confirm the classification of each source under Standards / Industrial / Peer-reviewed / Effective is defensible.
- [x] Reviewer decides whether to fold the follow-up wire-up of the new iterator terms into `statements.rst` / `expressions.rst` into this PR or a follow-up.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


gazprea/spec/glossary.rst
```diff
@@ -0,0 +1,876 @@
...
+   compilation.
+
+.. [#iso-26514] ISO/IEC/IEEE 26514:2022, *Systems and software
+   engineering -- Design and development of information for users*.
+   https://www.iso.org/standard/77451.html (also
+   https://standards.ieee.org/ieee/26514/7467/).  Year: 2022.  The
+   current revision of the standard covering the design and
+   development of user documentation, including guidance on
+   terminology and glossary sections; the front matter is available
+   free, the full text is paywalled.
```
Sir-NoChill: These should be kept as agent documentation, rather than placed in the actual glossary.

gazprea/spec/glossary.rst
```diff
@@ -0,0 +1,876 @@
...
+      :term:`value categories <value category>`.
+
+
+.. _ssec:glossary_authoritative:
+
+Authoritative sources
+---------------------
+
+The primary citations for the entries above are listed here.  All URLs
+were captured for later ``archive.org`` snapshotting.
```
Sir-NoChill: unnecessary note, delete.

gazprea/spec/glossary.rst
```diff
@@ -0,0 +1,876 @@
...
+
+.. [#ada-rm-5-5] ISO/IEC 8652:2012, clause 5.5 (Loop Statements),
+   paragraphs 6-9.  Defines the *loop parameter* of a ``for`` loop and
+   its iteration scheme.
+   http://www.ada-auth.org/standards/12rm/html/RM-5-5.html.  Year:
+   2012.
+
+.. [#ada-rm-annex-n] ISO/IEC 8652:2012, Annex N (Glossary of Terms),
+   entries "Elaboration" and "Execution".
+   http://www.ada-auth.org/standards/12rm/html/RM-N.html.  Year: 2012.
```
Sir-NoChill: Can we collapse references to a single specification to a single entry and list the specification as the generic citation? Then we can have 'sub' citations that reference specific sections. I think this would make the citation list more manageable.

gazprea/spec/glossary.rst
```diff
@@ -0,0 +1,876 @@
...
+   [defns.well.formed]: "C++ program constructed according to the
+   syntax rules, diagnosable semantic rules, and the one-definition
+   rule".  Draft mirror:
+   https://timsong-cpp.github.io/cppwp/n4868/defns.well.formed.  Year:
+   2020.
+
+.. [#cpp-defns-ill] ISO/IEC 14882:2020, definition [defns.ill.formed]:
+   "program that is not well-formed".  Draft mirror:
+   https://timsong-cpp.github.io/cppwp/n4868/defns.ill.formed.  Year:
+   2020.
```
Sir-NoChill: Same here, we should collapse to a CPP ISO 14882 citation and then note the sub-citations.

gazprea/spec/glossary.rst
```diff
@@ -0,0 +1,876 @@
...
+      [#cppref-value-cat]_.
+
+   well-formed
+      "C++ program constructed according to the syntax rules,
+      diagnosable semantic rules, and the one-definition rule"
+      [#cpp-defns-well]_.  *Gazprea* uses "well-formed" throughout in
+      this generalised sense: a *Gazprea* program is well-formed if it
+      satisfies every diagnosable rule stated in this specification.
+
+   xvalue
```
Sir-NoChill, line 579: for all of these value types, we can reference the cpp reference.

gazprea/spec/glossary.rst
```diff
@@ -0,0 +1,876 @@
...
+      this is a *terminological convention*, not an assertion that
+      *Gazprea*'s qualifiers behave like C's.
+
+   type system
+      A tractable syntactic method for classifying phrases of a
+      language by the kinds of values they compute
+      [#pierce-tapl]_.  Type systems are traditionally divided into
+      *static* systems (checking performed before execution) and
+      *dynamic* systems (checking performed during execution); see
+      Pierce, Chapter 1 [#pierce-tapl]_.
```
Sir-NoChill, line 538: This is good. We can also point to some of the interesting literature surrounding the curry-howard correspondance since I suspect there will be students interested in that.

gazprea/spec/glossary.rst
```diff
@@ -0,0 +1,876 @@
...
+      the identifier occurs" [#iso-c11]_ rather than the term "binding".
+      The academic origin of "binding" as the association of names with
+      denotable values in an environment is Strachey
+      [#strachey-2000]_.
+
+   object
+      "A region of data storage in the execution environment, the
+      contents of which can represent values" [#iso-c11]_.  In this
+      glossary the word always refers to the *run-time* entity, never
+      to an object in the sense of object-oriented programming.
```
Sir-NoChill: I think this is somewhat misleading due to the recent addition of the `vector` type to gazprea. This is likely a clarification we should make since the `vector` 'class' has methods and behaves very similarly to an object. We might need a discussion about this.

gazprea/spec/glossary.rst
```diff
@@ -0,0 +1,876 @@
...
+      A program that does not produce a target program, but "appears to
+      directly execute the operations specified in the source program on
+      inputs supplied by the user" [#dragon]_.  Contrast with
+      :term:`compiler` and :term:`translator`.
+
+   lifetime
+      "The portion of program execution during which storage is
+      guaranteed to be reserved" for an :term:`object`.  "An object
+      exists, has a constant address, and retains its last-stored value
+      throughout its lifetime" [#iso-c11]_.
```
Sir-NoChill, line 303: Maybe we should also reference the rust definition of the lifetime, given that the devs have done a very precise job of defining lifetimes for the borrow-checker.

gazprea/spec/glossary.rst
```diff
@@ -0,0 +1,876 @@
...
+      A program is *ill-formed* if it is not :term:`well-formed`
+      [#cpp-defns-ill]_.  A conforming implementation is required to
+      issue a diagnostic for at least one violation of a diagnosable
+      rule in an ill-formed program.
+
+   implementation-defined behavior
+      "Unspecified behavior where each implementation documents how the
+      choice is made" [#iso-c11]_.  Distinct from
+      :term:`unspecified behavior` (no documentation obligation) and
+      :term:`undefined behavior` (no requirements at all).
```
Sir-NoChill, line 259: We should note that gazprea should not have any 'user-distinguishable' impelmentation-defined behaviour or ill-formed programs. All programs should either be well-formed or emit an error.

gazprea/spec/glossary.rst
```diff
@@ -0,0 +1,876 @@
...
+
+   re-initialization
+      In a *Gazprea*
+      :ref:`iterator loop <sssec:statements_iter_loop>`, the binding,
+      performed at the start of every iteration, of the
+      :term:`iterator variable` to the next element of the captured
+      :term:`domain`.  Because re-initialization introduces a fresh
+      binding from the captured domain, neither reassignment of the
+      loop domain expression nor mutation of the iterator variable
+      inside the body carries information into the next iteration.
```
Sir-NoChill, line 224: These do not happen only in loops. This is a general term that describes the second (or nth) declaration of the same identifier within a given scope. We should clarify this.

gazprea/spec/glossary.rst
```diff
@@ -0,0 +1,876 @@
...
+   iterator variable
+      In a *Gazprea* :ref:`iterator loop <sssec:statements_iter_loop>`
+      or :term:`domain expression`, the identifier to the left of
+      ``in``.  A fresh binding is introduced at the start of every
+      iteration (see :term:`re-initialization`) and destroyed at the
+      end of that iteration.  Some *Gazprea* prose calls this the
+      *domain variable*; the two names denote the same entity.  In
+      other languages the equivalent binding is Ada's *loop parameter*
+      [#ada-rm-5-5]_ and C++'s *for-range-declaration*
+      [#cpp-draft-stmt-ranged]_.
```
Sir-NoChill: We should collapse this into the `iterator` variable, rather than the domain variable. Iterator is the more precise term and updates to the documentation should reflect this (probably inside generator definitions as well).

gazprea/spec/glossary.rst
```diff
@@ -0,0 +1,876 @@
...
+         ... Elaboration, the static phase, determines whether it is
+         well-typed and well-formed in other ways, and records relevant
+         type or form information in the basis" [#sml-defn]_.
+
+      When the size of a *Gazprea* array is fixed at the first
+      evaluation of its declaration rather than at :term:`compile time`,
+      we say the size is fixed at *elaboration time*, in the same sense
+      as Ada and VHDL.  Note that SML uses "elaboration" for a static
+      (compile-time) phase; *Gazprea*'s usage follows the Ada/VHDL
+      dynamic sense.
```
Sir-NoChill: Since we can only find two procedural languages (and rarely used ones at that) we might want to collapse this into some other definition like dynamic and point to alternative terms in that definition. Elaboration, in this case and for the purposes of gazprea, refers to the _first_ evaluation of a statement, which is rather unfortunate, but something we have to live with.

gazprea/spec/glossary.rst
```diff
@@ -0,0 +1,876 @@
...
+
+Every entry below has a **primary citation** to an authoritative source:
+an ISO/IEC or IEEE standard, the documentation of an ongoing industrial
+open-source project (LLVM, GCC, GNU Binutils), or a peer-reviewed
+publication in a respected venue.  Where a term has a widely-used *effective*
+reference (e.g. cppreference for the C++ value categories), that reference
+appears alongside the authoritative citation and is explicitly labelled as
+non-normative.  Rejected sources -- those we considered and did not adopt
+-- are listed in the :ref:`ssec:glossary_rejected` appendix so that a
+reviewer can audit our choices.
```
Sir-NoChill: I don't think we need to reference rejected sources (see the comment on that section).

gazprea/spec/functions.rst
```diff
@@ -100,9 +101,9 @@ function is encountered.
            }
          }
 
-``f`` is invalid since if ``b == false``, then we reach the end of the
-function without a return statement, so we don’t know what value
-``f(false)`` should take on.
+``f`` is :term:`ill-formed` since if ``b == false``, then we reach the
```
Sir-NoChill, line 104: We should note an example error that a conforming implementation should throw.

gazprea/spec/expressions.rst
```diff
@@ -49,12 +50,14 @@ Generators
...
-A generator consists of either one or two domain expressions,
-and an additional  expression on the right hand side of the bar (``|``).
+A generator creates a value of a 1D array type when one
+:term:`iterator variable` is used, and a 2D array type when two
+iterator variables are used.
+Any other number of iterator variables will yield an error.
+
+A generator consists of either one or two
+:term:`domain expressions <domain expression>`, and an additional
+expression on the right hand side of the bar (``|``).
```
Sir-NoChill, line 70: We make no mention of n-d iterator variables. I suppose this is not actually in the spec yet, but can you make sure? If we do mention n-d iterators or arrays, then we should amend this. Do we mention anywhere if vectors can be used as an iterator? This might need a separate clarification.

gazprea/spec/declarations.rst
```diff
@@ -21,11 +21,12 @@ omitted the default is ``const``, i.e. variables are immutable by default.
...
-to ensure functional purity. If the variables are not initialized to a known
-value their initial value might change depending on when the program is run.
-Therefore, if the explicit inialization is omitted it is equivalent to setting
-the value to zero.
+In *Gazprea* all variables must be initialized in a well defined manner in
+order to ensure :term:`functional purity`. If the variables are not
+initialized to a known value their initial value might change depending on
+when the program is run.
+Therefore, if the explicit inialization is omitted it is equivalent to
+setting the value to zero.
```
Sir-NoChill: hmm. This implies some definition of null. We should point to where the declaration documentation defines the default value for variables and note the term RAII for variable declaration.

gazprea/conf.py
```diff
@@ -77,3 +77,26 @@
...
+    # cppreference.com: 403 to bots.
+    r'^https?://en\.cppreference\.com/',
+    # The Word Factory: connection times out to bots.
+    r'^https?://(www\.)?thewordfactory\.com/',
+]
+
+# Give slow-responding but legitimate hosts more time before treating a
+# response as a failure.
+linkcheck_timeout = 30
+linkcheck_retries = 2
```
Sir-NoChill, line 130: I suppose this is a ci question, but we might want to use the sphinx build-in linkchecker instead of lychee, seeing as it already does what we need it to do.

.github/workflows/linkcheck.yml
```diff
@@ -15,8 +15,32 @@ jobs:
...
+        run: pip install -r requirements.txt
+
+      # Sphinx's built-in linkcheck builder honours the per-project
+      # linkcheck_ignore / linkcheck_anchors / linkcheck_timeout /
+      # linkcheck_retries entries in each conf.py, and treats redirects
+      # and rate-limited hosts more gracefully than lychee.  Run it
+      # against every Sphinx project in the tree; the builder itself
+      # exits non-zero when it finds a broken link, so we do not pass
+      # -W (which would also fail on pre-existing unrelated warnings
+      # like missing TODO entries).
```
Sir-NoChill: clean this to two lines, we don't need a paragraph.

gazprea/spec/glossary.rst
```diff
@@ -0,0 +1,867 @@
...
+      the first-time processing of a declaration whose effect --
+      allocating storage, computing size expressions, resolving
+      instantiations -- has to happen *when the declaration is
+      reached*, not at :term:`compile time`.
+
+      *Gazprea meaning.*  When the size of a *Gazprea* array is fixed
+      at the first evaluation of its declaration rather than at
+      compile time, the *Gazprea* spec says the size is fixed at
+      *elaboration time* -- meaning specifically "the moment the
+      declaration is first evaluated".
```
Sir-NoChill: I think we might want to use the subsequent fold into runtime as our definition for array sizes. We can note explicitly, maybe making reference to this as a parenthised definition, that the array size is initialized upon first execution of the definition at runtime and is immutable.

Sir-NoChill (commented):
Looking better, just a couple remaining comments to address and then I will mark this as an actual PR.

rcunrau (approved):
Looks amazing! Thanks! 

While I agree the glossary should be pointed at by the Gazprea spec, I think the other assignments would also benefit from linking to it.

I would also like to point at a stand-alone version from Canvas. Is that possible?

PR #110, opened 2026-08-08
[stack 1/20] chore(agents): scaffold reproducible session bootstrap

Sir-NoChill:
**PR 1 of the spec-review stack.** Infra only — no spec content changes. This one lands before the fix/refactor/spec PRs so the machinery that reproduces the review is in-tree from the start.

## What

Adds a new top-level `.agents/` directory containing a manifest-driven scaffold for reproducible agent-run sessions:

- `manifest.yaml` — declarative desired state (apt packages, commands with version probes, workspace skills, sentinel files)
- `bootstrap.sh` — idempotent apt installer + per-session agent GPG signing key generation
- `check.sh` — dependency validator (dry-run, no side effects)
- `healthcheck.sh` — drift/corruption detector for the scaffolded environment
- `render.py` — regenerates the three shell scripts from their `.tmpl` siblings against `manifest.yaml`
- `agent-pubkey.asc` — this session's exported public key (also handed to the user for forge registration)
- `README.md` update + `.gitignore` entries for volatile paths (`state/`, `gnupg/`, `agent-env.sh`)

## Why

Enables cross-session reproducibility of the spec-review workflow that produces the rest of this stack. Any subsequent agent (or human) picking up the repo can run `.agents/bootstrap.sh` and get an identical review environment, with an identity that is auditably separate from the human maintainer's commits.

## Vendor-neutrality

The scaffold does not name any particular agent tool:

- Directory is `.agents/`, not `.<vendor>/`.
- Default identity in `manifest.yaml` and `bootstrap.sh` is the generic `Agent`.
- The skills-directory path in `healthcheck.sh.tmpl` is a manifest variable (`skills_dir`), not a hardcoded `.<vendor>/skills/`.

## What is NOT in this PR

- No spec content changes (`gazprea/spec/*` untouched).
- No CI changes (Sphinx build and PR #70 linkchecker unaffected — nothing added to their scope).
- No new pip dependencies.

## Signatures

Commit signed with the agent's ephemeral session key (`F38D8669…FAC880`); public key vendored at `.agents/agent-pubkey.asc` and available for GitHub GPG verification.


.agents/agent-pubkey.asc
Sir-NoChill, line 1: Delete, we should not prescribe identity.

.agents/bootstrap.sh
```diff
@@ -0,0 +1,122 @@
...
+else
+  log "apt dependencies already satisfied"
+fi
+
+# ------------------------------------------------------- ephemeral signing key
+# Session key in a workspace-local GNUPGHOME [gnupg-unattended]; consumed by
+# `git agent-commit` via the AGENT_* environment (see agent-commit skill).
+AGENT_GNUPGHOME="${AGENTS_DIR}/gnupg"
+AGENT_GIT_NAME="Agent"
+AGENT_GIT_EMAIL="ai@blobfish.icu"
```
Sir-NoChill: This is not useful, other users may want their agents to sign commits differently. We should make this generic across users, maybe adding signing details to the manifest

.agents/bootstrap.sh
```diff
@@ -0,0 +1,122 @@
...
+expiry_epoch="$(gpg --homedir "${AGENT_GNUPGHOME}" --list-keys --with-colons "${fpr}" \
+  | awk -F: '$1=="pub" {print $7; exit}')"
+log "key ${fpr} expires $(date -u -d "@${expiry_epoch}" '+%Y-%m-%d' 2>/dev/null || echo 'never')"
+log "public key exported to ${AGENTS_DIR}/agent-pubkey.asc -- provide it to the user for forge registration [github-gpg]"
+
+cat > "${AGENTS_DIR}/agent-env.sh" <<EOF
+export AGENT_GIT_NAME="${AGENT_GIT_NAME}"
+export AGENT_GIT_EMAIL="${AGENT_GIT_EMAIL}"
+export AGENT_SIGNING_KEY="${fpr}"
+export AGENT_GNUPGHOME="${AGENT_GNUPGHOME}"
```
Sir-NoChill: See the above comment, we need to make this generic.

.agents/bootstrap.sh.tmpl
Sir-NoChill, line 1: Templating should be user agnostic. We can clean this up so that the boostrapping, render and setup scripts contain no 'identity' constructs, so that users can cleanly specify how and under what identity they want their agents to work.

.agents/bootstrap.sh.tmpl
```diff
@@ -0,0 +1,110 @@
...
+expiry_epoch="$(gpg --homedir "${AGENT_GNUPGHOME}" --list-keys --with-colons "${fpr}" \
+  | awk -F: '$1=="pub" {print $7; exit}')"
+log "key ${fpr} expires $(date -u -d "@${expiry_epoch}" '+%Y-%m-%d' 2>/dev/null || echo 'never')"
+log "public key exported to ${AGENTS_DIR}/agent-pubkey.asc -- provide it to the user for forge registration [github-gpg]"
+
+cat > "${AGENTS_DIR}/agent-env.sh" <<EOF
+export AGENT_GIT_NAME="${AGENT_GIT_NAME}"
+export AGENT_GIT_EMAIL="${AGENT_GIT_EMAIL}"
+export AGENT_SIGNING_KEY="${fpr}"
+export AGENT_GNUPGHOME="${AGENT_GNUPGHOME}"
```
Sir-NoChill, line 89: Don't set the identity here, leave that to the users.

.agents/bootstrap.sh.tmpl
```diff
@@ -0,0 +1,110 @@
...
+  log "apt dependencies already satisfied"
+fi
+
+# ------------------------------------------------------- ephemeral signing key
+# Session key in a workspace-local GNUPGHOME [gnupg-unattended]; consumed by
+# `git agent-commit` via the AGENT_* environment (see agent-commit skill).
+AGENT_GNUPGHOME="${AGENTS_DIR}/gnupg"
+AGENT_GIT_NAME="{{ agent.name }}"
+AGENT_GIT_EMAIL="{{ agent.email }}"
+AGENT_UID="${AGENT_GIT_NAME} <${AGENT_GIT_EMAIL}>"
```
Sir-NoChill: Do not specify identity.

.agents/check.sh
```diff
@@ -0,0 +1,77 @@
+#!/usr/bin/env bash
+# 415-docs dependency validation -- rendered from check.sh.tmpl.
+# Edit .agents/manifest.yaml and re-render; do not edit this file directly.
```
Sir-NoChill: If this is generated, I might argue that we should not commit it (likewise with bootstrap.sh)

.agents/check.sh
```diff
@@ -0,0 +1,77 @@
...
+  source "${AGENTS_DIR}/agent-env.sh"
+  if gpg --homedir "${AGENT_GNUPGHOME}" --list-secret-keys --with-colons "${AGENT_SIGNING_KEY}" 2>/dev/null \
+       | awk -F: '$1=="sec" && $2!="e" && $2!="r" {found=1} END {exit !found}'; then
+    ok "agent signing key ${AGENT_SIGNING_KEY}"
+  else
+    flag "agent signing key ${AGENT_SIGNING_KEY} missing, expired, or revoked"
+  fi
+else
+  flag "agent-env.sh missing; run .agents/bootstrap.sh"
+fi
```
Sir-NoChill: Don't prescribe identity.

.agents/check.sh.tmpl
```diff
@@ -0,0 +1,47 @@
...
+# Agent identity produced by bootstrap.sh and required by git agent-commit
+if [[ -r "${AGENTS_DIR}/agent-env.sh" ]]; then
+  # shellcheck source=/dev/null
+  source "${AGENTS_DIR}/agent-env.sh"
+  if gpg --homedir "${AGENT_GNUPGHOME}" --list-secret-keys --with-colons "${AGENT_SIGNING_KEY}" 2>/dev/null \
+       | awk -F: '$1=="sec" && $2!="e" && $2!="r" {found=1} END {exit !found}'; then
+    ok "agent signing key ${AGENT_SIGNING_KEY}"
+  else
+    flag "agent signing key ${AGENT_SIGNING_KEY} missing, expired, or revoked"
+  fi
```
Sir-NoChill, line 40: Don't prescribe identity, I think this is something that the user can set up themselves.

.agents/healthcheck.sh
```diff
@@ -0,0 +1,101 @@
+#!/usr/bin/env bash
+# 415-docs environment healthcheck -- rendered from healthcheck.sh.tmpl.
```
Sir-NoChill: Generated, don't commit. If it can be easily rebuilt from the template then we can generate it on setup with the render script.

.agents/healthcheck.sh.tmpl
```diff
@@ -0,0 +1,85 @@
...
+fi
+{% endfor %}
+
+# Agent key: expiry is the designed staleness alarm
+if [[ -r "${AGENTS_DIR}/agent-env.sh" ]]; then
+  # shellcheck source=/dev/null
+  source "${AGENTS_DIR}/agent-env.sh"
+  gpg --homedir "${AGENT_GNUPGHOME}" --list-secret-keys --with-colons "${AGENT_SIGNING_KEY}" 2>/dev/null \
+    | awk -F: '$1=="sec" && $2!="e" && $2!="r" {found=1} END {exit !found}' \
+    || drift "agent signing key unusable (likely expired); re-run bootstrap to mint a fresh one"
```
Sir-NoChill, line 71: Once again, identity prescription

.agents/manifest.yaml
```diff
@@ -0,0 +1,55 @@
...
+apt_packages:
+  - name: gnupg
+    version: null
+  - name: python3-yaml
+    version: null
+  - name: graphviz
+    version: null
+  - name: python3-pip
+    version: null
+
```
Sir-NoChill, line 25: We should prefer dependency management using `uv`. You can build a pyproject.toml for the base repository and then set up these bootstrapping tools to install uv, and then uv sync on a uv-managed venv.

.agents/manifest.yaml
```diff
@@ -0,0 +1,55 @@
...
+agent:
+  name: Agent
+  email: ai@blobfish.icu
+  key_expiry: 1w
+
+# skills_dir: on-disk location this session's agent tool loads skills from.
+# Leave unset when 'skills:' is empty. Populate both together when you want
+# the healthcheck to notice a workspace skill copy going missing.
+skills_dir: ""
```
Sir-NoChill: Probably useful to make this a list and to keep the repository `.agents/skills` directory in as a base skill.

.agents/render.py
```diff
@@ -0,0 +1,230 @@
+#!/usr/bin/env python3
+"""Render a jinja-subset template against a YAML manifest.
+
+Template syntax (a strict subset of Jinja semantics [jinja]):
```
Sir-NoChill: Any reason we're not just using jinja through the python package (jinja2)?

.agents/render.py
Sir-NoChill, line 1: We should use the pre-built and tested jinja renderer (https://pypi.org/project/Jinja2/)

.gitignore
Sir-NoChill, line 1: This is good, we can also provide a gitignore entry for a `.agents/scratch` directory just for local data.

README.md
```diff
@@ -9,3 +9,14 @@ The site is automatically updated using a Github Action when
...
+## Agent sessions
+
+At the start of every agent session:
+
+    .agents/healthcheck.sh && .agents/bootstrap.sh && source .agents/agent-env.sh
+
+After changing `.agents/manifest.yaml`: re-render from the `.agents/*.tmpl`
+templates (see `render.py`) and re-run bootstrap. A freshly minted agent key
+(bootstrap logs "generating", not "reusing") must be re-registered on the
+forge; the exported public key lives at `.agents/agent-pubkey.asc`.
```
Sir-NoChill: If this is the case, we delete all of the rendered files (and probably gitignore them as well) and then specify the re-generation of the files at the beginning of each session.

.agents/skills/grammar-consistency/SKILL.md
Sir-NoChill, line 1: Hmm, this is not what I expected. We should focus on english grammar (spelling, passive voice, subject consistency, technical writing quality) rather than the gazprea language grammar. The students are expected to implement the gazprea language from the quasi informal examples, obtaining clarification for syntax from a provided solution compiler when unclear. This skill is focused on providing the EBNF grammar for the gazprea language, which is part of the exercise for students. Rewrite this skill to focus on english language consistency.

.agents/skills/spec-review/check-ci.sh
Sir-NoChill, line 1: Can we just use a tool like `act` to run the actions locally rather than attempting to build a local harness via bash?

.agents/skills/spec-review/SKILL.md
```diff
@@ -0,0 +1,150 @@
+---
+name: spec-review
+description: Systematic editorial and structural review of one or more Sphinx source files in `gazprea/spec/`. Use this skill whenever you are asked to review, audit, sanity-check, or "read through" spec content in this repository -- including PR review over spec changes, pre-merge checks on a feature branch, or a fresh pass over an existing chapter. It defines the checklist a careful maintainer applies: heading hierarchy, cross-reference (`:ref:`, `:term:`, `:doc:`) integrity, glossary term coverage, admonition usage, RST directive correctness, unresolved TODO/FIXME/XXX markers, and a `gazc`-backed sanity check on inline Gazprea code blocks. Do NOT use this skill for grammar-fragment cross-consistency (that is [[grammar-consistency]]) or for glossary entry sourcing (that is the workflow in the memory `gazprea-glossary-source-audit`).
```
Sir-NoChill: See the rewriting the grammar-consistency skill, that actually should be used in conjunction with this skill after it is rewritten. If there is a way to codify how the current doc structure mandates the style of english prose we should attempt to implement that.

.agents/bootstrap.sh.tmpl
```diff
@@ -0,0 +1,135 @@
...
+
+# Sync the project venv from pyproject.toml. --frozen would require a
+# checked-in uv.lock; we allow a soft sync so contributors don't have to
+# regenerate the lock on every dependency bump [uv-sync].
+(cd "${AGENTS_DIR}/.." && uv sync)
+
+# ------------------------------------------------------- ephemeral signing key
+# Identity is optional. Bootstrap only mints a GPG key when the manifest's
+# `agent:` block names an identity; otherwise signing is left to the user's
+# own git configuration.
```
Sir-NoChill: Thank you, but we can make all of these comments much smaller. Note that this might just need to be part of a docker image we supply for development. We can evaluate that especially since agents are typically running in sandboxed environments anyways.

.agents/healthcheck.sh.tmpl
Sir-NoChill, line 1: Once again, I feel like all of this is probably better specified as a docker container. We might just prefer to write a ci-dev image in the cmput415/ci-utils image repo so that we don't need to set up or check the environment. We could even add a nix flake beside.

.agents/manifest.yaml
```diff
@@ -0,0 +1,61 @@
+# Render context for .agents/*.tmpl and desired state for healthcheck.sh.
+# This file is committed; identity is intentionally left blank. Each user
+# fills in an `agent:` block locally (or leaves it empty to skip signing).
+
+project: 415-docs
```
Sir-NoChill, line 6: we might want to point to a docker container here as well. We should make a PR to [this repo](https://github.com/cmput415/ci-utils.git) to build a docker container with all of our necessary utilities.

.agents/render.py
Sir-NoChill, line 1: This is perfect.

README.md
```diff
@@ -9,3 +9,33 @@ The site is automatically updated using a Github Action when
...
+blank by default; bootstrap only mints a GPG signing key when you fill it
+in with a name and email. Filling it in is a per-user choice -- treat the
+blank template as the shared committed state and keep your populated copy
+local. When you do configure signing, register the exported public key
+(`gpg --homedir .agents/gnupg --armor --export <fpr>`) on the forge before
+your first push.
+
+Bundled skills for spec review/consistency work live under
+`.agents/skills/` and are listed in `manifest.yaml`'s `skills:`. The
+healthcheck confirms each listed skill still has a non-empty `SKILL.md`.
```
Sir-NoChill, line 65: Let's modify to point to a dockerfile, then we need to keep this up to date.

Issue #132, opened 2026-08-17
[F26] Rubric Meeting


Sir-NoChill:
# Review and discussion points from August 17 Rubric Meeting

- Running both the compilers _and_ the generated code through valgrind - Coherence?
- We need to look inside people's code to see what they understand
- Should we drop the code points in favour of peer-review points?
  - @novo52 the lab exams work as a verification step, so that they serve as a 'gate' to ensure that the students know 
  - Can we set a minimum passing grade on the lab exams while keeping them at a small weighting?
- For each assignment we need a portion of the total weight
  - for each assignment we then assign a non-negligible portion to the lab exams
  - Generator - lab exam = 60%
  - LOLCode - lab exam = 50%
  - VCalc - lab exam = 33-40%
  - Gazprea - lab exam = 25-33%, @jnamaral has a different weighting perspective
- We need to budget TA time, but could we followup with the groups? The students need to know what they get assessed on. Should we talk with the english department?
- Get automatic class hierarchy diagrams from doxygen or jetbrains tooling
- remember 'cairo, no buildings are finished because the tax goes up when it is finished' - if you ask for a metric, you get the metric
- How do we tie in the project management tooling to this term
- Non-negligible amount of points for signing in questions
- Marking budget for the subjective portion 

## Action Items
- [ ] Writing up the deliverables for the oral exam - how does it affect the other marking? What happens if it's not one? @rcunrau 
- [ ] Writing up the deiverables for the interviewers @novo52 
- [ ] Take a look at how to build rubrics for the software engineering portions
- [ ] We should look at hours estimates for interviews with implementation teams
- [ ] Closing issues on the spec, namely #106 and vector related issues @Sir-NoChill 
- [ ] LOLCode spec tests currently take input, segment into multiple tests @rcunrau 
- [ ] Get it into the syllabus @rcunrau 
- [ ] Development Log Rules with a peer-evaluation within teams @Sir-NoChill 
- [ ] Make us teachers @rcunrau 
- [ ] How are the implementation details ( #103 ), should we merge them into the rubric?
  - [ ] Two new sections under grading for interviews and lab exams @novo52

novo52:
Suggestion for an alternative to interviews with teams: we could create channels in discord for each team to use for communication, which the TAs have access to. Then, we can see if they are going off the rails and if they are doing design work with each other. IDK if they should be enforced as the primary method of communication or if this is a good idea, LMK what you think.

Sir-NoChill:
[...]

Issue #101, opened 2026-07-17
Clarifications on ragged literals


novo52:
https://discord.com/channels/1392396664299458601/1446720730166202420

This discord thread wins the most confused and frustrated students award.

How do we resolve this?

The main problem here seems to be that vector literal syntax is the same as array literal syntax, so nested array literals can either be `vector<integer[*]>`, or `integer[*][*]`. A ragged literal then means different things depending on the context. Additionally, it is ambiguous whether arrays are implicitly promoted to vectors, so what exactly are the objects on the right-hand side of the examples given by students?

Sir-NoChill:
I think we have that in #106 , I'll fold it into the actual PR

PR #139, opened 2026-08-21
spec(gazprea): consolidated spec-review stack — nested types + audit fixes

Sir-NoChill:
# Consolidated spec-review stack — 20 PRs in one

This collapses the stacked-PR chain (`gh stack #121`, formerly `[stack 1/20]`–`[stack 20/20]`, PRs **#110–#138**) into a single PR against `master`. The 20 individual PRs are **closed**; their branches and comment threads stay readable at the links in the manifest below. **All 40 commits are preserved (nothing squashed)** — use the *Files changed* tab for the cumulative diff, or the commit list for the logical steps.

**Footprint:** 35 files, +1576 / −229. Spec: 24 files under `gazprea/spec/`. Infra: the `.agents/` reproducibility scaffold + `pyproject.toml`/`uv.lock` (from #110).

**Provenance:** grew out of the vector-semantics discussion (#106) and the Aug-17 rubric-meeting action item (#132). The final nested-types change closes **#106, #82, #71, #101, #86**.

---

## ⚠️ Cross-PR interactions — review the NET result, not each step

The stack intentionally states a rule and later revises it. Review the final tree, not the intermediate commits:

- **Vector element type: restricted (#113) → reversed (#138).** #113 added "`vector<T>` is a base type or 1-D array only; no vectors/tuples/structs/strings." #138 **reverses** this to "any storable type, nested to any depth." Net: permissive.
- **Struct/tuple nesting: example trimmed to obey the ban (#127) → ban reversed (#138).** #127 removed the `s1 struct_field` example to comply with the old "no struct-in-struct" rule; #138 **lifts** that rule (`struct.rst:10`, `tuple.rst:6`). Net: nesting legal; the example simply no longer demonstrates it.
- **Array rank: 2-D matrix ceiling → arbitrary rank (#138).** `T[n1]...[nk]`, any element type.
- **Live consequence for #124:** its flagged question "what about *nested tuple* / string / matrix members?" is now **live**, because #138 makes nested tuple/struct members legal.

---

## Open questions flagged for reviewer attention (per former PR)

These are the ambiguities each PR flagged as needing a language-owner decision (verify against `gazc` / the reference compiler where noted).

**#122 — combined struct decl form:** does `struct S (...) x;` accept a qualifier (`var struct S (...) x;`)? The "combined form is always `const`" claim is *inferred* from examples; if `gazc` accepts `var`, it's wrong.

**#123 — mixed concatenation result:** is `char || char` even legal? (patch's "(or characters)" asserts `char||char = char[]`.) Is the rule "at least one operand is `string`" or "either operand"? `char[] || char[] = char[]` is unexemplified.

**#124 — tuple member conversions:** members "convert by their kind" lists only scalar + array. **Nested `tuple`, `string`, `matrix` members are unaddressed** — now relevant post-#138. Confirm array-in-tuple casts pad/truncate like standalone array casts.

**#125 — uninit const & stride:** `by` is restricted to `>= 1` (rules out `by -1` reverse iteration — intended?). Uninit `const` is defined as "holds the zero value *permanently*" — confirm that vs. treating uninit-const as a type error.

**#126 — slice bounds:** confirm `..-i` selects `1..n-i`; confirm the 1-indexed assertion doesn't contradict another chapter; confirm grammar accepts `a[i..j] by 2` (vs `a[i..j by 2]`).

**#127 — struct fixed-length field:** patch chose `character[256]` over `string[256]`. If sized/bounded strings are intended, the fix differs. (See cross-PR note — the nesting ban it obeyed is now reversed.)

**#128 — procedure call sites:** are casts legal on a procedure-call result? (patch adds "and casts" to the operator whitelist.) Confirm the restatement's disallowed-site list is the exact inverse of the chapter-top allowed list.

**#129 — method-call surface (largest `feat:`):** (a) receivers are *variables* only — does `gazc` accept `x.len()` on a literal/expression result? (b) mutation limited to *function-local* vars vs. "no global/captured mutation" — which framing? (c) `append` tie-break = single-element wins — matches reference? (d) "method call is the *only* expression usable as a statement" — reconcile with `call` statements. (e) **`concat`→`append` rename is breaking** — does the reference still expose `concat` as an alias?

**#114 — range vs array-index:** is **matrix slicing** intended? (`matrix.rst:103` still "indices must be integers"). No formal def of "range syntax" vs "range expression"; `by`-on-slice rule lives in a different chapter and may drift.

**#115 — promotion vs cast:** `type_casting.rst:48` still says "promoted" for an explicit `as<>` cast; `type_casting.rst:95` allows array→array casts across dimensions — confirm 1-D↔2-D explicit cast is intended.

**#138 — nested aggregates & n-d arrays (deferred, not in scope):** a rank-agnostic `shape` interface (vs `rows`/`columns`, per #82), n-d matrix-multiply, and broadcasting are left to a follow-up. The ragged (`vector<vector<T>>`) vs rectangular (`vector<T[*]>`, `T[*][*]`) split from #101 is implied but not yet spelled out operationally.

**Lower-priority / follow-ups noted in-PR:** #116 (`string.rst:8` loose equivalence); #118 (`integer.rst` Associativity column duplication); #111 (`expressions.rst:54` generator-error class; repo-wide "must emit" sweep); #112 (prose `String`/`Vector` capitalization left per maintainer decision). #119, #120, #130 flagged no open questions.

---

## Infra (`.agents/` scaffold, from #110) — open feedback digest

PR #110 drew 25 inline comments (@Sir-NoChill). Themes (several already addressed by later commits on the same branch — `223c45d` *drop identity prescription & generated artifacts*, `6e2bbba` *point .agents/ at the DocsDoc image + rewrite skills*, `66ae45f` *uv lockfile*):

- **Don't prescribe identity** — signing identity should be user-supplied via the manifest, not hardcoded; delete the vendored `agent-pubkey.asc`. *(→ addressed by `223c45d`.)*
- **Don't commit generated files** — `bootstrap.sh`/`check.sh`/`healthcheck.sh` are rendered from templates; gitignore them and regenerate on setup. *(→ addressed by `223c45d`.)*
- **Prefer a Docker/CI image** — make this a dev image in `cmput415/ci-utils` (optionally a nix flake); point the manifest/README at the container instead of a bash env harness. *(→ partly addressed by `6e2bbba`.)*
- **Dependency management via `uv`** — `pyproject.toml` + `uv sync` on a uv-managed venv. *(→ addressed by `66ae45f` + pyproject.)*
- **Templating via `jinja2`** — use the jinja2 package rather than a custom renderer (`render.py` was later approved: "This is perfect").
- **Rewrite the `grammar-consistency` skill** — focus on **English prose quality** (spelling, passive voice, subject consistency, technical writing), *not* the Gazprea EBNF (that's the students' exercise); the `spec-review` skill should consume it after the rewrite. *(→ addressed by `6e2bbba`.)*
- **Local CI via `act`** rather than a hand-rolled `check-ci.sh` bash harness; add a `.agents/scratch` gitignore entry.

Full thread: #110.

---

## Manifest — the 20 collapsed PRs

| # | Title | Branch |
|---|-------|--------|
| #110 | chore(agents): scaffold reproducible session bootstrap | `chore/agents-scaffold` |
| #112 | spec: replace undefined syntax in examples | `fix/undefined-syntax-examples` |
| #113 | spec: state the vector element-type set *(reversed by #138)* | `fix/vector-element-types` |
| #116 | spec: scope the vector-array equivalence claim | `fix/vector-array-divergence` |
| #114 | spec: distinguish range slices from array-valued indices | `fix/range-index-semantics` |
| #115 | spec: scope the promotion-implies-cast claim | `fix/promotion-cast-claim` |
| #118 | spec: single home for operator precedence | `refactor/precedence-single-home` |
| #119 | spec: single home for stream_state semantics | `refactor/stream-state-home` |
| #120 | spec: single home for const-by-default rule | `refactor/const-default-home` |
| #111 | spec: name error classes at anonymous error sites | `feat/errors-chapter` |
| #122 | spec: specify combined struct-plus-instance declaration | `fix/struct-decl-form` |
| #123 | spec: state the mixed concatenation result type | `fix/string-concat-result` |
| #124 | spec: tuple member conversions cover array members | `fix/tuple-conversion-members` |
| #125 | spec: close two silent contract gaps (uninit const, stride) | `fix/uninit-const-and-stride` |
| #126 | spec: unify slice bound rule in one section | `fix/slice-bounds` |
| #127 | spec: struct field / fixed-length field *(nesting ban reversed by #138)* | `fix/struct-nesting` |
| #128 | spec: reconcile procedure call-site restatement | `fix/procedure-call-sites` |
| #129 | spec: specify the method-call surface *(largest feat)* | `feat/method-calls` |
| #130 | spec: correct misleading examples (batched) | `fix/example-corrections` |
| #138 | spec: permit nested aggregates and n-d arrays | `feat/nested-composite-types` |

<sub>Consolidated from `gh stack #121`. Individual PRs closed in favour of this one; commit history preserved.</sub>


rcunrau (approved):
Let's do it!

novo52:
We may have to look into the three namespaces more. I think the example conflates the struct namespace (unsure if gazprea has this, but I think C does, for example) with the struct field namespace (gazprea and C have this)

novo52:
The glossary and declarations page disagree on the zero-value of a character. (`\0` vs `' '`). I prefer `\0`.

novo52:
We still need to apply the inclusive-exclusive range-slice parity from #109

novo52:
I am fearful of `const` slices being able to change their values when the underlying array is modified, but I'm not sure there is any real harm/danger in this.

novo52:
There is a problem with the vector ragedness explanation what will need resolution. 

As it stands it seems like it is saying that if a nested array literal is being assigned to a vector, different padding rules apply to the vector than if it was being assigned to an array variable. I think this is not good.

We should be consistent that any expression like `[[], []]` is an array literal and always has the same behaviour—I think padding to the largest element is good. The example should show that if you assign `const vector<real[*]> x = [[1.0], [2.0, 3.0]]` it will pad the first element, as the array literal is what is being padded. In contrast, `var vector<real[*]> y = [[1.0, 2.0]]; y.push([3.0])` will pad the new entry to match length two. These two padding rules are not clear in the spec, or, reading the spec as it stands, are not sensible.

novo52:
Other than what I've said above, this is great.

rcunrau:
I thought neither of these are allowed/legal, which I suppose points out that the spec needs to be clearer.

For the first one: `[[1.0], [2.0,3.0]]` I thought the first element determined the size of all elements that followed, i.e. this is a size/type error. If the array were `[[1.0,2.0],[3.0]]` we would pad the second element.

For the second example, I thought we (I) decided *not* to pad. Making radded right vectors of vectors is legal and caveat emptor.

gazprea/spec/types/array.rst
```diff
@@ -269,10 +269,24 @@ Operations
...
 
 
+      Concatenation is right-associative, and its *receiver* -- the rightmost
+      operand -- fixes the **kind** of the result. When the receiver is a
+      :ref:`vector <ssec:vector>` (a ``vector<T>`` or a ``string``), the whole
+      concatenation is a vector of that element type; otherwise -- when the
+      receiver is an array or a scalar -- the result is an array, exactly as in
+      the examples above. Nothing else about concatenation changes: at least one
+      operand must still be composite, and the operands must share a common
+      element type through implicit casts. This is what keeps a string
```
novo52: I thought the thing that would make x a string is that it is a string variable and the assignment will implicit cast. I think with that in mind, this receiver formulation is unnecessary.

gazprea/spec/types/array.rst
```diff
@@ -506,48 +523,67 @@ element:
...
-An array slice is a **view** into the elements of its backing array, not a
-copy:
+Whether a slice copies or writes through depends on where it appears: a slice is
+a **copy when read** and a **view when assigned to**.
 
--  As an :term:`lvalue`, a slice writes through to its backing array, and
-   so is an lvalue only when that array is mutable (declared ``var``):
+-  **In value position** (an :term:`rvalue`) -- as an initializer, on the right
+   of an assignment, as an argument bound to a ``const`` parameter, or anywhere
+   an array value is expected -- a slice produces a **fresh, independent array**
```
novo52: This is wrong according to Ron’s intention for slices

gazprea/spec/types/array.rst
```diff
@@ -506,48 +523,67 @@ element:
...
    (``a[1..3]`` selects indices 1 and 2.)
 
--  As an :term:`rvalue`, a slice is a live, read-only view. A ``const``
-   slice is such a read-only view and need not be built on a ``const``
-   array -- it still reflects later changes to its backing array:
+-  **In assignment-target position** (an :term:`lvalue`) -- on the *left* of an
+   assignment, or bound to a ``var`` reference parameter -- a slice is a **view**
+   that writes *through* to its backing array. This is the only situation in
+   which a slice aliases storage, and it requires the backing array to be mutable
+   (declared ``var``); a slice of a ``const`` array is never an lvalue. The
```
novo52: Should be the same error as assignment to const—AssignError, I believe

gazprea/spec/types/array.rst
```diff
@@ -561,11 +597,52 @@ A slice whose (in-bounds) left bound is greater than its right bound, such as
...
+
+    // axis 1 by an integer (dropped); axes 2 and 3 by ranges (kept):
+    var b = a[1][1..3][1..3];   // a fresh integer[2][2] copy == [[1, 2], [4, 5]]
+
+    // the same positional selection as an lvalue writes through to a:
+    a[2][1..3][1..3] = [[28, 29], [30, 31]];  // updates those four elements of a
+
+Because a subscript chain names successive axes rather than re-indexing an
+intermediate result, writing more index positions than the array has axes is
+*not* how one indexes into a slice's result; for that, bind the slice to a
```
novo52: The same operator now has two different meanings

gazprea/spec/types/array.rst
```diff
@@ -561,11 +597,52 @@ A slice whose (in-bounds) left bound is greater than its right bound, such as
...
+variable (or parenthesize it) and index the resulting value, as shown above.
+
+A slice may also be handed to a :ref:`function <sec:function>` or
+:ref:`procedure <sec:procedure>`. In an argument position it follows the same
+copy-or-view rule as everywhere else, decided by the *parameter* it binds to:
+
+-  A slice bound to a ``const`` parameter is passed **by value** -- the callee
```
novo52: As discussed, saying it is passed by value is misleading—in most implementations it would be passed by reference as usual, but the compiler assures that it is never written to

novo52:
The slice changes are a step back in my opinion. They were good as they were in the last meeting with the exception of the fact that slices cannot be assigned to a variable and thus are never const or var in themselves. That has been correctly removed, but most of the added stuff is not good.

Multidimensional slices breaks a property of array and vector representation. Previously, arrays, vectors, and slices are all represented by {ptr, len} and all operations (indexing in particular) compose in the obvious way. Now, indexing by ranges gives non-obvious behaviour and slices cannot be represented as {ptr, len}.

The copy vs read distinction is not needed. If we assume a slice is always a view that lasts for the duration of an expression, then all properties fall out naturally. A slice ends up copied as an r value not because the slice copies, but because assignment always copies—uniform with other array assignment semantics.

I would recommend reverting the slices commit and just deleting the Const vs var paragraph that implies slices can be assigned to a variable as views.

novo52:
Rather than having both slices and ranges be half open, they should both be closed—this is the way most existing 1-indexed languages treat ranges, as this formulation is more natural for 1-indexing.

Push back if you disagree.

gazprea/spec/declarations.rst
```diff
@@ -54,32 +54,25 @@ This applies to ``const`` declarations as well: a ``const`` variable
 declared without an initializer is legal and holds the zero value of
 its type permanently.
 
-For simplicity *Gazprea* assumes that declarations can only appear at
-the beginning of a block. For instance this would not be legal in
-*Gazprea*:
+A declaration may appear at **any** point within a block; *Gazprea* does not
+require the declarations of a block to be grouped at its start, so a declaration
+may be interleaved freely with the statements around it. For instance, this is
+legal even though a declaration follows an ordinary statement:
```
novo52: This language should not assume so strongly that students would assume declarations must be grouped. This is ordinary behaviour for a programming language.

gazprea/spec/types/array.rst
Sir-NoChill: The way I have it specified in vectors is that vector-ness is never propagated, since the vector is _always_ used as an array in arithmetic expressions. The reciever formulation allows for strings to print normally if the reciever is a string. I have string specified as a true compiler-provided typealias for a char array and that printing is the only difference.

So under my vector rules, if we wanted to assign a string to a variable, it would require a type definition.

## Before this rule

```gaz
var a = "ab" || ['c', 'd'] || "de";

a -> std_output; // prints "[a b c d e]"
"ab" || ['c', 'd'] || "de" -> std_output; // prints "[a b c d e]"
```

After type resolution, I'm thinking that RHS is treated like a char array _for the sake of assignment and expression resolution_ since a vector is _never_ part of an expression and implicitly two-way castable to a the equivalent array with size fixed at the vector's current length.

Under this rule it becomes unwieldly to print raw characters/strings to the terminal.

I want to do it this way because the vector is not a true 'primitive type' and so the language can have an escape hatch here.

## After this rule

```gaz
var a = "ab" || ['c', 'd'] || "de";

a -> std_output; // prints "[a b c d e]"
"ab" || ['c', 'd'] || "de" -> std_output; // prints "abcde"
```

Since the reciever dictates the output formatting, then the string prints nicely.

## Alternatives

### Vector-ness is Propagated Everywhere

We could change it so that vector-ness is propagated through, but then do vectors downcast to arrays in expressions or do arrays upcast and what becomes the rules on size errors via expressions?

I think that answering the size error questions in the context of vectors becomes significantly harder than answering them via arrays like we currently do.

### Assignment propagates the Widest Type

So vectors cross the boundary to the assignment rather than being elided to their underlying array.

I am open to suggestions here, but having any single element being a vector and that being propagated changes the implications of expressions always using their underlying array as the arithmetic type; or at least results in a bit of an arbitrary contradiction.

gazprea/spec/types/array.rst
Sir-NoChill: This is one that I should have flagged, I have been considering mutability part of the type.

I think my argument is grounded in gazprea, since const and var parameters have different semantics and historically the compiler implementations decide mutability at type check.

I think that I would rather remove the assign error in favour of having this and other examples fail type check instead, but I can revert this if that is the vibe.

gazprea/spec/types/array.rst
Sir-NoChill: Yes, it is, but slices on the LHS are not really slices, they are a 'splat'.

Building subarrays from indexes I think is a natural extension of using slices to begin with. I have defined the definition of new variables from splats as being a truly 'new' variable, so a copy. If you hate this, I can revert, but I have a further defense in my subsequent comment.

gazprea/spec/types/array.rst
Sir-NoChill: consequence of n-d arrays I think. We could change it so that n-d arrays are instead written and indexed in the form `<array>[axis, axis, axis][index]`, which I think is cleaner, but requires a big change. Could also go the numpy way and just keep the indexing as positional.

gazprea/spec/types/array.rst
Sir-NoChill: I believe I clarify this in the next sentence or two, that being 'passed by value' is more of a conceptual way to think about const arrays. Also in MLIR this abstraction works, we do actually pass the full 'tensor' or whatever we are using as an array, rather than as a reference. The bufferization pass handles the reference/value semantics if done correctly. Likewise for vars, in mlir everything is pass by value and functionally returned without side effects and the students can abuse this feature to help with some of their analysis, especially since globals are all immutable.

gazprea/spec/declarations.rst
Sir-NoChill: yes, agreed.

gazprea/spec/types/array.rst
novo52: The root of the problem I think is that indexing is a rank reducing operation, while slicing is a rank preserving operation. Therefore, indexing chains naturally to specify a position in the nd array as a series of indices, while composing slices just repeatedly slices.

If we were to add multidimensional slices we should go the numpy way, but I think we should keep them out of the spec until we see how students handle/implement slices this year.

gazprea/spec/types/array.rst
novo52: I agree, your following explanation is good. I think though that the surrounding lvalue/rvalue view/copy language should be removed.

gazprea/spec/types/array.rst
novo52: As long as it is consistent, either option is good for me.

Sir-NoChill:
I withdraw my claim of having a well-reasoned argument for extending slice semantics, but if that's the case then are we allowed to slice an n-d array? Or should we remove all n-d array capabilities and go back to rank-2 maximum?

novo52:
I think slicing an nd array is fine, but it only slices the outermost extent, in the natural way assuming slices are 1d.

in the future I think we can extend to numpy like syntax for nd slicing.

Sir-NoChill:
> The slice changes are a step back in my opinion. They were good as they were in the last meeting with the exception of the fact that slices cannot be assigned to a variable and thus are never const or var in themselves. That has been correctly removed, but most of the added stuff is not good.

This resolves the slices being assignable, fundamentally a different operation, more like a 'splat' than a slice. Of course we could add a new 'splat' operator for this use case.

> Multidimensional slices breaks a property of array and vector representation. Previously, arrays, vectors, and slices are all represented by {ptr, len} and all operations (indexing in particular) compose in the obvious way. Now, indexing by ranges gives non-obvious behaviour and slices cannot be represented as {ptr, len}.

So we disallow slicing on any non-terminal rank, does that mean we can also not slice vectors containing arrays?

My argument for this was that we can slice them, we can even assign to n-d arrays, and it is all still computable via index operations from the (ptr, len) tuple. It actually works nicely if you implement slicing as an indexing modifier rather than an 'operation' in itself, then it just becomes a mapping.

We do not have the issue of this being a reference to the underlying array since _slices_ are not themselves storable, so I'm not sure I see the issue with slicing the n-d arrays here.
 
> The copy vs read distinction is not needed. If we assume a slice is always a view that lasts for the duration of an expression, then all properties fall out naturally. A slice ends up copied as an r value not because the slice copies, but because assignment always copies—uniform with other array assignment semantics.

If we remove the splatting, then yes. I like splats because it makes it easy to build an array out of a different, potentially modified array or vector where we do not necessarily know the element we want, but if removing them seems like the better option then I will cede.

> I would recommend reverting the slices commit and just deleting the Const vs var paragraph that implies slices can be assigned to a variable as views.

This should already be deleted, I have it specified that slices can be assigned via the splat which just coincidentally shares view syntax but on the LHS. Could be confusing, so could make a new operator, but also could wholesale delete any mention of slices on the LHS or in expressions, which is the implication. Under deletion slices can _only_ be an R-value and can _never_ be used in an expression context.

gazprea/spec/types/array.rst
novo52: I think our understanding of the language semantics is aligned, but we disagree on how it is presented. Assigning the result of a slice to a variable produces a copy because assignment always copies.

I think of it like an extension of indexing or Tuple element selection

```
var arr = [1, 2, 3];
arr[1..2] = [10, 20]; // LHS is a lvalue, the is an rvalue
const x = arr[2..3]; // LHS is an lvalue, the slice is an lvalue
// semantically the same as if I do
const y = arr; // arr is an lvalue, assignment copies the array
```

the way you explain it also works with this example, but it is more complex to reason about.

novo52:
Would you like to call to discuss the splat stuff?

rcunrau:
I didn't think we ever allowed multi-dimensional slicing, so sorry if I missed that whole discussion. The prime reason I wanted slicing is to allow Vectors to interoperate with arrays. We have concluded that n-d vectors can never interoperate with n-d arrays, so I think there's not much point in trying to figure out n-d slices. 

The only thing that comes to mind is that an n-d array is still <pointer, len=N1*N2*...Nn>, so a slice could still be mapped onto the array. In that sense a slice is range indexing, and I can't remember if we still have that or not.

rcunrau:
WRT splatting, I'm not at all interested in extending the language at this point.

Sir-NoChill:
consider it dead

Sir-NoChill:
Alright, I think I got most of the issues hammered out. I made a couple editorial decisions:

1. I allowed use array slicing like array literals in expressions, since they are always const and can always be described as a pointer and length
2. I allowed typealiases inside scopes other than the global scope to be consistent with structs
3. Structs are nominal types
4. I removed mentions of the argument passing conventions, as students will be using MLIR, not the raw llvm ptr-style passing style.

I am, once again, terribly sorry for the +5000 line PR, so feel free to identify my failings gratuitously.

rcunrau:
I think we should merge it and go from there

Sir-NoChill:
Merging. We'll fix it in prod :laughing:

PR #140, opened 2026-08-25
Work on lab machines

novo52:
Recommend the lab machines for all students, and expand on the details of how to work on them.


rcunrau (approved):
What toolchain do we currently have on the lab machines? Do we need to update the LLVM version and also the documentation about it?

